PR #1, opened 2021-03-14
Update README.md

alok8bb:
This is how I would like it to be ;)
Change things you don't like (like those `-` on headers)


alok8bb:
BrUH

siduck76:
Thanks for the correction.

PR #6, opened 2021-03-19
add install script

jack0son:
Script to automatically symlink config, install packer, and chosen language servers.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+BASE=$(git rev-parse --show-toplevel)
+LSP_BIN_PATH=$HOME/.local/bin
+
+default_langs="ts rust"
+lsp_langs=${@:-"$default_langs"}
+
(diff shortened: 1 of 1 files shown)


siduck76:
> Script to automatically symlink config, install packer, and chosen language servers.

since the config  supports html , css , the install script must have it mentioned under default langs

jack0son:
> > Script to automatically symlink config, install packer, and chosen language servers.
> 
> since the config supports html and css , the install script must have it mentioned under default langs .

The default langs is for lsp dependencies rather than treesitter. Could add a treesitter config section as well.

siduck76:
treesitter config is already provided so no need for that and besides html and css lsp dependencies are also required so it must be added .

jack0son:
> treesitter config is already provided so no need for that and besides html and css lsp dependencies are also required so it must be added .

Okay makes sense.

PR #10, opened 2021-03-31
Add support for the python language server

marvelman3284:
I mainly code in python so I added support for the python language server (pyls) in my personally dots, thought it wouldn't harm if everyone had access to them.


siduck76:
@marvelman3284 does it work? test the install script and isnt Jedi language server better? It would be nice if the install.sh script prompted users to select lsp servers they want .

PR #14, opened 2021-04-02
install.sh for init use and pip3

aserowy:
- added -p to mkdir to ensure the dir is created on clean installs
- added pip3 to pip validation


siduck76 (approved):
Thanks for the fixes :)

PR #23, opened 2021-04-06
added nvim-lspinstall and removed manual lsps

aserowy:


aserowy:
here we go :)

if you are missing languages, feel free to ping me. Thus i'll contribute those to the lspinstall repo (currently linux only).

kind regards
Alexander

siduck76:
Are there any default languages that'd get install by default rather than using :LspInstall to do  install them? @aserowy

aserowy:
not right now. tell me the defaults you want and i tinker them in and commit to this pr.

siduck76:
Html , css , js (tsserver) , python , rust

aserowy:
bad news everyone.. it seems packager has an open issue resolving around auto installing plugins is not possible without using the ui atm. Thus, i cannot fire commands in headless mode.
The only option would be to hardcode the languages as defaults.

Is this requirement mandatory for you? @siduck76

siduck76:
@aserowy I think yes , cant we just run :LspInstall html css  etc  without opening nvim? like adding it in the install.sh

siduck76:
> @aserowy I think yes , it would be nice if we had some default languages supported , cant we just run :LspInstall html css etc without opening nvim? like adding it in the install.sh

aserowy:
hi,

sorry for the long pause. work and family :D
like i said above, we cannot run nvim in headless, because we cannot install packages on init by now. thus, we cant call commands on lspinstall cause it isnt installed automatically.

I will close the pr for now.

kind regards
alexander

siduck76:
@aserowy lsp install is finally added now :)

PR #27, opened 2021-04-14
aliases on the top

Aaditay:
Why not make an alias of cmd first and then use cmd instead of vim.cmd


siduck76 (approved):
@Aaditay Thanks for the correction.

PR #31, opened 2021-04-24
use gnu sed in mac

mTvare6:
Normal said gives `command a expects \ followed by text` error `gsed` which is part of gnu coreutils for macOS works


siduck76 (commented):
Thanks for adding this , I havent tested it on Mac yet . hopefully it works for you :)

PR #36, opened 2021-05-05
Fix a smaller error that prevents the changing of shells

marvelman3284:
With the changing of file names and file structure the ability to select your own shell didn't work.


siduck76:
I totally forgot about it. Thanks for this!

PR #40, opened 2021-05-07
Bug fix when executing any packer command and Add Clang LSP

SergioRibera:


siduck76:
@SergioRibera clangd alone wont be enough I think , It would be better to add ccls too!

siduck76 (approved):

SergioRibera:
> @SergioRibera clangd alone wont be enough I think , It would be better to add ccls too!

You are right, for the next pull request I will try to implement it.

siduck76:
no worries , I already added it :)

PR #42, opened 2021-05-09
Update README.md

Vanderscycle:
The Readme keys were not matching the defined keys in top-bufferline.lua.

Also, I tried your keys, but kept the old ones (shift + l | p I think) because shift + , | . doesn't work for me. the shift key changes them from , | . into < | > 

I rebinded on my machine so it works for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,9 +208,15 @@ I'd install it first and add its setup line :
 
 - **Managing tabs with bufferline!**
 
+<<<<<<< HEAD
 - <kbd> Shift </kbd> <kbd> l or p </kbd> cycle through opened tabs
 - <kbd> Shift </kbd> <kbd> x </kbd> close current tab
 - <kbd> Shift </kbd> <kbd> t </kbd> open new tab
+=======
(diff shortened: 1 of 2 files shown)


siduck76:
@Vanderscycle yes I recently got to know that , and . cant be used for mappings .Can you include in the PR ,  those bufferline cycling mappings that work for you too?

Vanderscycle:
Changed the binding back to S-l | S-p and made the change in the README.md I wonder if C-, |C-. is a better binding that matches your initial intend.

siduck76:
I dont think its possible to map anything with , or .  , and S-p / S-l are already some of vanilla vim's default keybinds so I dont think we should use that. 

image

Vanderscycle:
Hey, I don't often revert commits so this is a bit of a learning experience. I think that the README didn't revert fully as its showing the git <<< head or ====

siduck76:
@Vanderscycle thats okay , I'll edit your commits . Im just confused which mappings to use for bufferlinecycling . Most mappings are already occupied already :(

siduck76 (approved):

PR #47, opened 2021-05-12
Addition of which key

Vanderscycle:
PR for the addition of which key.
Change to misc-util with the addition of a timeout time of 500 ms. whichkey.lua contains the default settings for which key and can be modified

reference:
https://github.com/folke/which-key.nvim

### DIFF
--- a/lua/misc-utils.lua
+++ b/lua/misc-utils.lua
@@ -23,6 +23,7 @@ opt("o", "cmdheight", 1)
 
 opt("o", "updatetime", 250) -- update interval for gitsigns
 opt("o", "clipboard", "unnamedplus")
+opt("o", "timeoutlen", 500)
 
 -- for indenline
 opt("b", "expandtab", true)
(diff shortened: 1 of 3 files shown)


siduck76:
its supposed to look like this by default? 

image

Vanderscycle:
Looks about right, idk why mine displays over 4 rows vice 2 
whichKey

Vanderscycle:
If you press <leader> for example and let it timeout, which-key will popout with only the <leader> bindings. Same with v and z. Bsc from any of the sub menus will bring you the "main" menu (the one you posted).

siduck:
@Vanderscycle it has been added recently, I assume you dont use nvchad but reminding it xD

siduck:
image
image

Vanderscycle:
Glad to hear, I moved to Lunarvim which gave me the time and space to dive deeper into nvim api. NvChad, at the time, was changing very quickly and I didn't have the time to keep up.

PR #60, opened 2021-06-15
Expanding LSPinstall capabilities

Vanderscycle:
Hey,

I began working with vue.js as my first framework and played with my config a bit more (I am still running an old version waiting for .5 to release). One of the element that I couldn't install using lspinstall was emmet which speeds up boilerplate writting.

I find it difficult at times customizing with lspinstall, and I was thinking about LunarVim and its language implementation. This pr is essentially a blueprint on how you can take from any available languages listed [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md) and deeply customize the behaviour of the lsp you are using. 

tl;dr: I think lunarVim does a really good job wrt lsp and customization.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -57,3 +57,8 @@ vim.api.nvim_exec([[
 require "zenmode"
 require "whichkey"
 require "dashboard"
+
+-- lsp (not found on lsp install) list bellow:
+-- https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md
+-- steps: 1) add a file in the lsp folder (kinda like lunar nvim); 2) add its requirement here
+require 'lsp.emmet'
--- a/lua/lsp/emmet.lua
+++ b/lua/lsp/emmet.lua
@@ -6,7 +6,7 @@ capabilities.textDocument.completion.completionItem.snippetSupport = true
 configs.emmet_ls = {
   default_config = {
     cmd = {'emmet-ls', '--stdio'};
-    filetypes = {'html', 'css'};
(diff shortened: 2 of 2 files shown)


siduck76:
I appreciate the PR for this! I think Lspinstall is already working on it https://github.com/kabouzeid/nvim-lspinstall/issues/34 , it would be better if you add a PR there it would be much better. Since emmet-ls config wouldnt be added unless the user installed it via lspinstall (:LspInstall emmet) and this would help the config from expanding more! 

you could check some examples of adding lsp servers in lspInstall , it looks easy : 

https://github.com/kabouzeid/nvim-lspinstall/blob/main/lua/lspinstall/servers.lua
https://github.com/kabouzeid/nvim-lspinstall/blob/main/lua/lspinstall/servers/css.lua

I could put your emmet config in the wiki if you're unable to add PR at lspInstall :)

Vanderscycle:
I didn't think about creating a pr for lspInstall. Should be safe to add a small guide to the wiki detailing how to install an lsp that is not available in lspInstall.

siduck76:
done adding it! https://github.com/siduck76/NvChad/wiki#--install--lsp-servers

Vanderscycle:
Thks,

Interresting, is there a way to open the wiki to contributions with review? You added a space by accident when inserting the lspconfig language url.

siduck76:
Im not sure about the contributions on wiki , you can check https://stackoverflow.com/questions/10642928/how-to-pull-request-a-wiki-page-on-github.

PR #66, opened 2021-06-15
Added commenter support.

ashincoder:
Added commenter support with `leader + / : CommentToggle`
Leader being the space key


siduck76 (approved):

ashincoder:
Yup actually that was better.

PR #74, opened 2021-06-16
Added relative numbers option and changed bookmark icon

ashincoder:
Things added:
- Added relative numbers option. It is currently commented in misc-utils.lua


ashincoder:
changed bookmark icon

ashincoder:
Removed repeated lines @siduck76

PR #82, opened 2021-06-17
Add developer plugins & settings

jtuz:


siduck76:
why vim polygot? we already have treesitter

siduck76:
indicator icon would make the tabs look bad , Im fine with not having it

ashincoder:
These settings looks like an overkill

siduck76:
@jtuz if some stuffs dont work with whichkey for you , mention them and give proper configs so I will at least put them in wiki so people facing same issue as yours will be helped too :)

PR #83, opened 2021-06-17
More comments and Edited Readme

ashincoder:
I added more comments for knowing.
I removed one screenshot.
I explained about Treesitter.

## TODO 
Change README and add:

- How to add custom plugins
- How to custom settings
- How to change themes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,6 @@ An attempt to make neovim (cli version) as functional as an IDE while being very
 # Screenshots -
 
 <img src ="https://raw.githubusercontent.com/siduck76/dotfiles/master/rice%20flex/dashboard-nvim.png"><hr>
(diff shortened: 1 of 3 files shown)


siduck76:
This feels unnecessary since those things dont need comments like 

require "statusline" already means adding statusline config and same with others .

ashincoder:
It doesn't affect I added those lines just for continuity.

ashincoder:
It is always better to add comments if not needed too. I mean it doesn't affect anyone.

siduck76:
> It is always better to add comments if not needed too. I mean it doesn't affect anyone.

yes but our module are named better like pluginlist , anyone would get to know that its the moduleof plugins list, we dont have to write comments everywhere!

PR #96, opened 2021-06-25
removed syntax highlight

ashincoder:
I commented the `syntax on` option. Since it is uneccessary because (treesitter is enable). It slows down the startup time too. Check it and see the difference

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -18,7 +18,7 @@ g.mapleader = " "
 g.auto_save = 0
 
 -- colorscheme related stuff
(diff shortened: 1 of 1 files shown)


siduck76:
@ashincoder I'll add this soon, I already decided many changes and commits for those 

image

like each of these folders have NvChad with major changes , I'll commit them one by one so it gets easier to update via diffs

siduck76:
@ashincoder It has been removed , thanks for this . And now this has made the startup more faster!

Issue #121, opened 2021-07-04
Path color in NvimTree


ashincoder:
**Describe the bug**
The path of directory is specified in NvimTree. But it is dispayed in dark text so we cannot see it.

**To Reproduce**
1. Open NvimTree
2. Go to the top
3. You will notice

**Expected behavior**
To show the path of the directory in light text.

**Screenshots**
It is displayed now because I selected the text
![2021-07-04-192123_1366x768_scrot](https://user-images.githubusercontent.com/83629316/124387661-0f82d080-dcfd-11eb-8664-d325f84908c6.png)

**Desktop (please complete the following information):**
 - Void Linux
 - St patched
 - 0.5

**Additional context**
Changing the color will solve the problem.

ashincoder:
This issue will be closed when PR #122 is merged.

se7en-x230:
I ./install.sh --update but I still have the problem. Arch, st terminal

![image](https://user-images.githubusercontent.com/69777713/126607501-455939b9-2f17-4510-b8aa-fb3fc1096ecb.png)

siduck76:
> This issue will be closed when PR #122 is merged.

It's made hidden because the cwd shows in the statusline

siduck76:
> I ./install.sh --update but I still have the problem. Arch, st terminal
> 
> ![image](https://user-images.githubusercontent.com/69777713/126607501-455939b9-2f17-4510-b8aa-fb3fc1096ecb.png)

Go to highlights.lua file and search for rootdir in nvim section, change highlight for it

se7en-x230:
Thanks

PR #122, opened 2021-07-04
Fixed some things

ashincoder:
Changes made:
 - Changed relative number format option to current way.
 - Fixed error #121


siduck76:
@ashincoder nvimtreefolder color has black color set because i want to hide it 
image

we have cwd displayed in statusline itself so revert that.

ashincoder:
Ok reverted it @siduck76

PR #124, opened 2021-07-04
Color settings

alternateved:
Clean up colorsheme-related stuff from `init.lua` and move it to separate file in `nvim/lua` folder.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -8,13 +8,9 @@ local g = vim.g
 g.mapleader = " "
 g.auto_save = false
 
--- colorscheme related stuff
+-- colorsheme
+require "colors"
 
-g.nvchad_theme = "onedark"
-local base16 = require "base16"
-base16(base16.themes["onedark"], true)
-
-require "highlights"
 require "mappings"
 require "file-icons"
 require "statusline"
--- a/lua/colors.lua
+++ b/lua/colors.lua
@@ -1,0 +1,17 @@
+local cmd = vim.cmd
+
+cmd "syntax on"
+local base16 = require "base16"
+
+vim.g.nvchad_theme = "onedark"
+base16(base16.themes["onedark"], true)
+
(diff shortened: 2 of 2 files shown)


ashincoder (changes requested):
maybe remove the cmd syntax on since it works without it and also it causes a delay in loading nvim

alternateved:
Right. The modification is done now.

siduck76:
@alternateved The init.lua's already very short , adding those 4-5 lines to a separate file would make the config look bigger. Also the theme is an important part of the config so I put it in the init.lua file , whenever we open the ~/.config/nvim config folder its init.lua can accessed directly unlike colors.lua which is in the lua folder ( using nvimtree for this might take more time than just opening init.lua)

ashincoder:
@siduck76 if we add `require "colors"` at the top of the document it will be loaded first. So no big deal

ashincoder:
Adding different files in the lua directory is more better right. Seperate config for seperate purpose.
**Unix Philosopy**

alternateved:
I guess it is a matter of preference. I prefer to have my `init.lua` file just for initiating everything else. Like this:
image

It is much more comfortable for me to just edit the file which contains specific functionality.

ashincoder:
yeah me too. That is the name of the file init.lua

siduck76:
> @siduck76 if we add `require "colors"` at the top of the document it will be loaded first. So no big deal

the order cant change , some plugins are loaded before the highlights etc , we might need to load certain plugins after it and its a big hassle lol so I kept it that way :p

siduck76:
Also the init.lua looking that empty doesnt look so right to me , yes ik this is very biased lol. I'll try to keep files as less as possible in nvchad and since init.lua isnt more than 100+ lines we'll define the theme in it.

PR #125, opened 2021-07-04
Add background colors in case of different terminal colors

alternateved:
Background colors for three themes in case terminal's background color is different.


siduck76:
use bg("Normal","hexcolor") also add it tomorrow night in the list too.

siduck76 (approved):

PR #130, opened 2021-07-05
Made some changes in mappings and settings

ashincoder:
Added:

- Made z prefix for zenmode mappings
- Added time out length setting in misc-utils.lua


siduck76:
@ashincoder why timeoutlen? and whats that

ashincoder:
Time out len is used for many cases. To fast up the lsp completion etc.
Anyone can change to their preference

PR #131, opened 2021-07-05
Update TrueZen.nvim configuration file

alternateved:
It seems that [TrueZen.nvim](https://github.com/Pocco81/TrueZen.nvim) options for configuration have been reworked. This is an updated file for the previous NvChad setup.


alternateved:
Reopening without unnecessary `plugin/packer_compiled.lua` file.

PR #133, opened 2021-07-05
Added parser for markdown

ashincoder:
Added parser for markdown in treesitter


siduck76:
btw why separate config for markdown ?

ashincoder:
idk why . But it works for markdown

ashincoder:
Yeah don't use it unless it is merged with nvim-treesitter

PR #134, opened 2021-07-05
added special telescope mapping for neovim config files

ashincoder:
added special telescope mapping for neovim config files
SPC f l


siduck76:
@ashincoder this works very well , And @Vanderscycle you dont need to make the PR anymore

siduck76:
btw can @ashincoder can we cd in the ~/.config/nvim folder via telescope too? if we would get this warning since the default cwd is home
 
image

ashincoder:
Oh I check it

ashincoder:
I don't think we can.
Users may prefer ~/

siduck76:
> I don't think we can.
> Users may prefer ~/

cd ~/.config/nvim/ | Telescope find_files cwd=~/.config/nvim

siduck76:
If we open nvim tree ( and it shows the home folder contents ) then it will not matter even if we did cd to any folder , nvimtree wont update its content. but if we didnt open nvimtree and cd into any folder , the nvimtree will update its content itself.
https://user-images.githubusercontent.com/59060246/124499267-5e069c80-dddb-11eb-8d2a-970dfd92c9cc.mp4

ashincoder:
yeah it works

ashincoder:
Shall I do a PR

siduck76:
> Shall I do a PR

no wait , we'll wait till this thing gets fixed https://github.com/kyazdani42/nvim-tree.lua/issues/489

ashincoder:
OK

siduck76:
> OK

no need to close it , re-open it if you can.

ashincoder:
Hello I dont't think the command I wrote was good, cause when I use which key it show's the whole command.

ashincoder:
check this https://github.com/folke/which-key.nvim/issues/119. I think we can use which key

PR #140, opened 2021-07-07
Added lazy loaded which key plugin

ashincoder:


siduck76:
ehhh not yet! I need to check whichkey configuring first , it gets hard for me if someone sends PR of something which I dont know.

siduck76:
So keep this PR opened for a while , I'll check whichkey next month.

ashincoder:
What about which key confuses you

siduck76:
> What about which key confuses you

its not about confusing , Im just now familiar with it much . I'll check its docs soon

PR #145, opened 2021-07-09
Refactor Install script | Fix a mapping bug

Akianonymus:
install.sh changes: As much as i like automated scripts, they are a very high risk high reward thing. For that reason, i have refactored the script to check for potential edgecases that can mess up our existing stuff, obviously it's far from perfect but somewhere i guess.

About the second commit, i am assuming that was not the intended behaviour.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,74 +1,200 @@
 #!/bin/sh
 
-# check if git command is installed
-command -v git > /dev/null || {
-  printf "Install git before proceeding\n"
-  exit 1
+_usage() {
+  printf "%s" \
+    "Usage: sh ${0##*/} [ options ]
+    -h | --help          => Show this help.
+
+    -i | --install       => Install the config.
+
+    -u | --update        => Update the existing config without removing existing stuff.
+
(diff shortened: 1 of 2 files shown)


siduck76:
> install.sh changes: As much as i like automated scripts, they are a very high risk high reward thing. For that reason, i have refactored the script to check for potential edgecases that can mess up our existing stuff, obviously it's far from perfect but somewhere i guess.
> 
> About the second commit, i am assuming that was not the intended behaviour.

bd! has already been added , shift + x ( under bufferline mappings) so I think its better to just remove the ctrl+q mapping.

siduck76 (approved):

Akianonymus:
> > install.sh changes: As much as i like automated scripts, they are a very high risk high reward thing. For that reason, i have refactored the script to check for potential edgecases that can mess up our existing stuff, obviously it's far from perfect but somewhere i guess.
> > About the second commit, i am assuming that was not the intended behaviour.
> 
> bd! has already been added , shift + x ( under bufferline mappings) so I think its better to just remove the ctrl+q mapping.

Ah, missed that somehow.

PR #147, opened 2021-07-10
Dev

ashincoder:


siduck76:
I still have to consider about this , maybe using a cheatsheet for keybinds would be better.

PR #160, opened 2021-07-17
Add escape to normal mode functionality with plugin

G-Rowell:
Hi all,

I've added the [better-escape.vim](https://github.com/jdhao/better-escape.vim) plugin with this pull request,
the aim of the plugin is to enable the common escape key-bind: `-- vim.cmd("inoremap jh <Esc>")` 
This key-bind is present (but disabled) [in NvChad already](https://github.com/siduck76/NvChad/blob/main/lua/mappings.lua#L38).

BUT with better UI performance as it doesn't wait to print the 'j' character to the buffer. 

**NOTE: very minimal impact to StartupTime**
(I measured with the plugin disabled and enabled, the difference was ~ 0.5sec)

**NOTE: [this is disabled by default](https://github.com/siduck76/NvChad/pull/160/commits/76b58daad054bd90b5ee230f03f81f551fa91fcb#diff-16c5fdf1ca1112f3bff35b8d14a6de8a174c71ee275a1f057dbd3f4fbbeeff83R142),** within: 'pluginList.lua' line 142: "disable = true,"
(disabled by default as the original key-bind was disabled by default)

**tl;dr** escape bind feels far less laggy, mokeh like, monkeh happy

### DIFF
--- a/lua/mappings.lua
+++ b/lua/mappings.lua
@@ -35,7 +35,6 @@ map("n", "<leader>zm", ":TZMinimalist<CR>", opt)
 map("n", "<leader>zf", ":TZFocus<CR>", opt)
 
 map("n", "<C-s>", ":w <CR>", opt)
--- vim.cmd("inoremap jh <Esc>")
 
 -- Commenter Keybinding
 map("n", "<leader>/", ":CommentToggle<CR>", opt)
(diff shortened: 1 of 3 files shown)


G-Rowell:
**StartupTime results**

with plugin: ~28.5
without plugin: ~ 28

without-escapeStartupTime
escapeStartupTime

siduck76:
adding plugin for this simple thing wouldnt be a good idea! 

just use 

vim.api.nvim_set_keymap("i", "jk", "<esc>", {})

PR #162, opened 2021-07-18
Fix terminal config issues

G-Rowell:
This is a pretty small commit/pull, hope that's okay

This fixes an escape mapping and terminal opening mapping

**Fixes:**
1. escape 'jk' mapping not working in terminal mode
2. escape 'jk' mapping not working in visual mode
3. opening a new terminal buffer (with `<CNTL> + t + t`) did not hide line numbers
         (only after viewing another buffer and returning)
         it did work for the `<CNTL> + l` & `<CNTL> + x` mappings, as they were just splits
         (it still works for these mappings)

Planning to investigate some more term config, but starting with this
         
tl;dr: can escape term & visual better, 
            monkeh still cannot escape the sad

### DIFF
--- a/lua/mappings.lua
+++ b/lua/mappings.lua
@@ -18,7 +18,10 @@ map("v", "x", [=[ "_x ]=], opt)
  this line too ]]
 --
 
+-- escape with 'jk' mapping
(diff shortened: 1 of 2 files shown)


siduck76:
2021-07-18-230027_1920x1080_scrot

PR #165, opened 2021-07-19
Even faster startup time | Fix formatting | Misc

Akianonymus:
install.sh: Run PackerSync from script after install too

```
as the corresponding code for that was removed from init.lua
```

init: Asynchronously load filetype and syntax

```
lots of improvement for startup time
```

Fix formatting | Misc

```
* fix formatting in pluginList, luasnip
 * don't use pcall on highlights
 * sync and compile fzf and media files only when the error occurs, add a
  help text telling tbe user to restart neovim
```
                                                   
packerInit: Clone packer at opt path as we are lazy loading

```
move packadd to packerInit
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -153,11 +153,7 @@ main() {
   if _NVIM="$(command -v nvim)"; then
     if _check_nvim_version; then
       printf "\n%s\n" "=> Neovim will now open." && sleep 1
-      if [ "${_UPDATE}" = "false" ]; then
-        "${_NVIM}" +":lua require 'pluginList' vim.cmd('PackerSync')"
-      else
-        "${_NVIM}"
-      fi
+      "${_NVIM}" +":lua require 'pluginList' vim.cmd('PackerSync')"
     else
       printf "Error: Neovim is installed, but version is lower than 0.5.x, install Neovim >= 5.x and then run nvim & do :PackerSync\n."
     fi
--- a/lua/options.lua
+++ b/lua/options.lua
@@ -19,7 +19,7 @@ opt.clipboard = "unnamedplus"
 opt.shortmess:append("sI")
 
 -- disable tilde on end of buffer: https://github.com/  neovim/neovim/pull/8546#issuecomment-643643758
-vim.cmd [[let &fcs='eob: ']]
+vim.cmd("let &fcs='eob: '")
 
 -- Numbers
 opt.number = true
--- a/lua/options.lua
+++ b/lua/options.lua
@@ -1,6 +1,13 @@
 local opt = vim.opt
 local g = vim.g
 
+-- Turn these off at startup, will be enabled later just before loading the theme
+vim.cmd([[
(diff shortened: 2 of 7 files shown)


Akianonymus:
Before:

Screenshot_20210719-064239_Termux_1


After:

Screenshot_20210719-064346_Termux_1

siduck76:
@Akianonymus  THANKS A LOTTTTTTTTTT  <3

Akianonymus:
> @Akianonymus THANKS A LOTTTTTTTTTT <3

Btw, the startuptime plugin is broken i think 😂.

We should use https://github.com/dstein64/vim-startuptime

I benchmarked with it

Screenshot_20210719-085609_Termux_1

It actually shows the time elapsed, not dummy stuff.

siduck76:
nah its not broken -.-- 

btw nvchad on ssd 

image

Akianonymus:
> nah its not broken -.--
> 
> btw nvchad on ssd
> 
> image

Yes it is, it's ridiculous that it loaded all the that short time, it the time to just load the init.lua file, not actually neovim.

If that's what you want to see, loading time of init.lua, then sure, it is showing the correct info.

Pocco81:
image

This is just ... wow .... My SSD isn't even the fastest and yet I'm here with this startup time.

Also @siduck76, you should consider disabling writing the `shada` file, that would decrease the startup time as well, however at this point idk if you want to get that even lower lol.

siduck76:
> image
> 
> This is just ... wow .... My SSD isn't even the fastest and yet I'm here with this startup time.
> 
> Also @siduck76, you should consider disabling writing the `shada` file, that would decrease the startup time as well, however at this point idk if you want to get that even lower lol.

send a lil PR xd

Akianonymus:
> Also @siduck76, you should consider disabling writing the `shada` file, that would decrease the startup time as well, however at this point idk if you want to get that even lower lol.

I actually tried it, not much difference.

And it is actually important

```The ShaDa file is used to store:
  - The command line history.
  - The search string history.
  - The input-line history.
  - Contents of non-empty registers.
  - Marks for several files.
  - File marks, pointing to locations in files.
  - Last search/substitute pattern (for 'n' and '&').
  - The buffer list.
  - Global variables.
```

siduck76:
> > Also @siduck76, you should consider disabling writing the `shada` file, that would decrease the startup time as well, however at this point idk if you want to get that even lower lol.
> 
> I actually tried it, not much difference.
> 
> And it is actually important
> 
> ```
>   - The command line history.
>   - The search string history.
>   - The input-line history.
>   - Contents of non-empty registers.
>   - Marks for several files.
>   - File marks, pointing to locations in files.
>   - Last search/substitute pattern (for 'n' and '&').
>   - The buffer list.
>   - Global variables.
> ```

yeah it removes telescope oldfiles feature too so nono!

PR #166, opened 2021-07-19
Fixes

Akianonymus:


siduck76:
The theme name should stay in the theme.lua , shouldnt get separated from it!

Akianonymus:
@siduck76 How about this, it will remain in theme.lua, but a fallback theme name in options.lua, just in case for first install ( ik, too much for first install, but it is what it is ).

siduck76:
> @siduck76 How about this, it will remain in theme.lua, but a fallback theme name in options.lua, just in case for first install ( ik, too much for first install, but it is what it is ).

i dont think its needed for first install . just add dark bg for NormalFloat cuz it looks light pink during the first install xd

Akianonymus:
@siduck76 Well, i will remove the theme stuff for now, gotta figure out something else.

siduck76 (approved):

PR #168, opened 2021-07-20
Dev test custom user config

G-Rowell:
Hi all,

[This is related to this issue.](https://github.com/siduck76/NvChad/issues/68)

We are making custom user configs for NvChad, which don't interfere/merge-conflict 

.gitignore won't work for this, as the file is comitted,
we use `git update-index --assume-unchanged lua/custom.lua`

This seems to be a working prototype, but I'm intending this to be a part of [the discussion here](https://github.com/siduck76/NvChad/issues/68)
So feel free to comment, not certain this is ready to merge (but it is working!)

**tl;dr:** monkey set their own config to be banana coloured

### DIFF
--- a/lua/user_config.lua
+++ b/lua/user_config.lua
@@ -1,0 +1,18 @@
+-- WARN: this file is specially ignored by git
+-- Src: https://stackoverflow.com/a/26245961
+-- to 'ignore' this file: git update-index --assume-unchanged <file>
+-- to un-'ignore' this file: git update-index --no-assume-unchanged <file>
+
+local function map(mode, lhs, rhs, opts)
+    local options = {noremap = true, silent = true}
+    if opts then
+        options = vim.tbl_extend("force", options, opts)
+    end
+    vim.api.nvim_set_keymap(mode, lhs, rhs, options)
+end
+
+local options = vim.opt
+local g = vim.g
+
+--I want to disable the line numbering for my setup
+--options.number = true


G-Rowell:
Need to check if `git update-index --assume-unchanged lua/custom.lua` is persistent across each user's own copy of the branch

Akianonymus:
How about adding a custom.sample.lua to repo, not custom.lua

But, we will originally use custom.lua only. A check will be added if custom.lua is present for error handling

custom.sample.lua will contain all the configuration which are change able without breaking with core nvchad setup.

By all configuration, i mean vim options( enable dashboard or not ), mappings, etc

The sample custom config will contain the values which are set in nvchad by default.

So, even if user renames custom.sample.lua to custom.lua, should get default configs and then they can start messing with it .

G-Rowell:
MMmm I like it defs some good ideas,
why not keep it as `custom.lua` but comment out the sourcing/calling of this file in `init.lua`?

Though could keep some commented options as example,
could also have a wiki page listing options (or list breaking options/plugins not tune from within `custom.lua`

Think you have some good points! :)

G-Rowell:
Note for future me: 
Seems like this may have an issue,
replicate:
modifiy custom.lua,
git status shows nothing,
git checkout main bring error due to custom.lua modify

Akianonymus:
> Note for future me:
> Seems like this may have an issue,
> replicate:
> modifiy custom.lua,
> git status shows nothing,
> git checkout main bring error due to custom.lua modify

One of the main reason i said custom.sample.lua was because of this, we won't to worry about it. And no need to overwriting user custom.lua because we don't have custom.lua in this repo to begin with .

G-Rowell:
> > Note for future me:
> > Seems like this may have an issue,
> > replicate:
> > modifiy custom.lua,
> > git status shows nothing,
> > git checkout main bring error due to custom.lua modify
> 
> One of the main reason i said custom.sample.lua was because of this, we won't to worry about it. And no need to overwriting user custom.lua because we don't have custom.lua in this repo to begin with .

Yeah, think you've given wiser input than me
Ty Ty, let's have a look sometime :smile:

G-Rowell:
Closing, this was a proof of concept and no longer needed.
There are some brilliant ideas from many contributors for a better implementation.

PR #171, opened 2021-07-20
Securely call modules and disable all builtin plugin

Pocco81:
This PR disables all builtin plugins because it wasn't being done before and also securely calls all of nvchad's modules.

**Why do we need to securely call the modules?**

The secure call feature that this PR implements assures that users will get a clear trackback to where the issue originated. This is an example of what the error message looks like if `x` given module fails to load:

image

As you can see it's easy to identify what's causing the issue.

Furthermore, another _huge_ advantage secure calls have is that no matter if the module being required failed, the other will still be able to work with "what's left". This guarantees that no matter how high up the issue occurs the rest should still be loaded. For example:

image

If there was an error loading the `plugins.bufferline` module, then neither the `mappings` module nor the `utils` module would be loaded. However, with the secure calls implementation they will still load regardless of `plugins.bufferline` succeeding or not. 

Edit: forgot to mention that you shouldn't try to load the `options` module inside of the `async func`. Of course you can also securely call that but it must be outside of the `asyc`

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -1,15 +1,24 @@
 require "options"
 
+local chad_modules = {
+    "pluginList",
+    "plugins.bufferline",
+	"mappings",
+	"utils",
+}
+
 local async
 async =
     vim.loop.new_async(
     vim.schedule_wrap(
         function()
-            require "pluginList"
-            require "plugins.bufferline"
-            require "mappings"
-            require("utils").hideStuff()
-
+            for i = 1, #chad_modules, 1 do
+                local ok, res = xpcall(require, debug.traceback, chad_modules[i])
+                if not (ok) then
+                    print("NvChad [E0]: There was an error loading the module '" .. chad_modules[i] .. "' -->")
+                    print(res) -- print stack traceback of the error
+                end
+            end
             async:close()
         end
     )
--- a/init.lua
+++ b/init.lua
@@ -15,7 +15,7 @@ async =
             for i = 1, #chad_modules, 1 do
                 local ok, res = xpcall(require, debug.traceback, chad_modules[i])
                 if not (ok) then
-                    print("NvChad [E0]: There was an error loading the module '" .. chad_modules[i] .. "' -->")
+                    print("Error loading module : " .. chad_modules[i])
                     print(res) -- print stack traceback of the error
                 end
             end
--- a/init.lua
+++ b/init.lua
@@ -3,8 +3,8 @@ require "options"
 local chad_modules = {
     "pluginList",
     "plugins.bufferline",
-	"mappings",
-	"utils",
+    "mappings",
+    "utils"
 }
 
 local async
--- a/lua/options.lua
+++ b/lua/options.lua
@@ -46,16 +46,29 @@ g.mapleader = " "
 g.auto_save = false
 
 -- disable builtin vim plugins
-g.loaded_gzip = 0
-g.loaded_tar = 0
-g.loaded_tarPlugin = 0
-g.loaded_zipPlugin = 0
-g.loaded_2html_plugin = 0
-g.loaded_netrw = 0
-g.loaded_netrwPlugin = 0
-g.loaded_matchit = 0
-g.loaded_matchparen = 0
-g.loaded_spec = 0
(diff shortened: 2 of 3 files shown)


Akianonymus:
Good catch but it won't even trigger, we are using individual pcalls everywhere require is called.

For example, about plugins.bufferline, even if it fails, it won't crash the hole config because we are already doing the error handling.

https://github.com/siduck76/NvChad/blob/665fa24783d4a6ebf513a05ced6b7c8e0eea3c24/lua/plugins/bufferline.lua#L4

```
local present, bufferline = pcall(require, "bufferline")
if not present then
    return
end
```

For testing my theory, just change that code to

```
local present, bufferline = pcall(require, "bufferline")
if present then
    return
end
```

See if mappings are loaded are not.

About options.lua, yes it should be outside, but we can move it inside except for some options like termguicolors, disabling the built in plugins, shortmess, tilde, etc.

Pocco81:
How come? Let me give you some examples:

Forgot to add ending bracket in a plugin at pluginsList.lua? gets triggered:

image

Forgot to add parenthesis, commas, etc... to mappings.lua? gets triggered:

image

`vim.g.nvchad_theme` [for some odd reason] doesn't exist when needed at bufferline.lua? gets triggered:

image


I can keep on going...

Akianonymus:
@Pocco81 Actually, that was not my point. Obviously what you said is correct.

But by that logic you have do everything under a protected call, like even in options.lua

vim.opt.something, if something doesn't exist, it will error too.

My point is, use protected call on unknown scenario, like a fetched plugin in packer. If are going to assume that even the local code present in the repo is faulty, then what's the point.

Anyway, it should be good if doesn't impact the startup time ( theoretically protected calls will take more computing power, dunno how relavant that is in real life usage ). On that note, handle require options too in a xpcall/pcall if we are going that way.

Akianonymus:
Also, fix the formatting of init.lua

Pocco81:
> But by that logic you have do everything under a protected call, like even in options.lua

No need for that. `xpcall()` catches the error and gives the stack traceback back. So there is not need to explicitly add `xpcall()` to everything.

> Also, fix the formatting of init.lua

My bad! I'm using `luafmt` with `neoformat` to format my lua code. Do I need to use a specific formatter or something?

Akianonymus:
> No need for that. `xpcall()` catches the error and gives the stack traceback back. So there is not need to explicitly add `xpcall()` to everything.

My bad, i actually meant, xpcall for options.lua, not inside.

> My bad! I'm using `luafmt` with `neoformat` to format my lua code. Do I need to use a specific formatter or something?

This one, https://github.com/trixnz/lua-fmt

You have to trigger formatting manually using leader+fm

> (I think I posted my startup time a while ago here. It was like 5.something back then because I disabled syntax and other stuff at startup, but that messed my colorscheme and diagnostics and other stuff, so I reverted it)

To be honest, you benchmarks are not to be considered, you have a sdd, xD. I will check on my potato laptop to see if startuptime is increased ( probably not ).

Pocco81:
> My bad, i actually meant, xpcall for options.lua, not inside.
Yeah maybe we could that. I'm actually doing that in my personal config (nvchad fork)

> This one, https://github.com/trixnz/lua-fmt
Yeah I'm also using that one, I just forgot to format it. Just added [this commit](https://github.com/siduck76/NvChad/pull/171/commits/5531edf8e46e4d805acab5075ec5bc628538e150) to fix it :)

> To be honest, you benchmarks are not to be considered, you have a sdd

Not only that lol. I benchmarked it manually so my data shouldn't be trusted lol.

Akianonymus:
> > My bad, i actually meant, xpcall for options.lua, not inside.
> 
> Yeah maybe we could do that. I'm actually doing that in my personal config (nvchad fork)
> 

Just

```
local ok, trace = xpcall(require, debug.traceback, "options")
if not (ok) then
    error("NvChad [E0]: There was an error loading the options.lua\n" .. trace)
end
```

Use error because it should proceed to parse the config further.

siduck76 (approved):

PR #178, opened 2021-07-23
install.sh: Use chsh instead of grep /etc/shells | init: Move bufferline config to pluginList | Misc improvements

Akianonymus:
mappings.lua: Code style fixes | Cleanup options.lua

    plugins.bufferline should be as config of nvim-bufferline

    install.sh: Use chsh instead of grep /etc/shells

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -2,7 +2,6 @@ require "options"
 
 local chad_modules = {
     "pluginList",
-    "plugins.bufferline",
     "mappings",
     "utils"
 }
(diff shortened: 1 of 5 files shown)


Akianonymus:
Btw, don't merge this yet, may have some edgecases ( i have yet to find ).

Akianonymus:
@siduck76 this can be merged now ( but do test in case ).

siduck76 (approved):
neat

PR #186, opened 2021-07-26
Use jdhao/better-escape.vim | Allow moving the cursor through wrapped lines normally

Akianonymus:
Use jdhao/better-escape.vim to handle jk as escape mappings

    only for insert mode ( so also works on terminal too )

    remove visual mode key binding, till we find a proper solution, because it is super annoying right now

    Original pr here: https://github.com/siduck76/NvChad/pull/160

---

mappings: Allow moving the cursor through wrapped lines normally

    This doesn't affect any other stuff like 10j or 10k

### DIFF
--- a/lua/mappings.lua
+++ b/lua/mappings.lua
@@ -18,11 +18,6 @@ map("v", "x", [=[ "_x ]=], opt)
  this line too ]]
 --
 
--- escape with 'jk' mapping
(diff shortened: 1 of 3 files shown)


Akianonymus:
Before you go on asking is the plugin worth it ? Yes it is totally.

siduck76 (approved):

PR #191, opened 2021-07-27
Mappings

Akianonymus:


siduck76:
I had added the file explorer text on nvim top but it doesnt look good , looks good and clean without it so I removed it. And this is much of user setting so It's "" by default.

Akianonymus:
Closing as commit moved to https://github.com/siduck76/NvChad/pull/193

PR #195, opened 2021-07-28
add XDG_CONFIG_HOME

mmatongo:
I tried installing it on my system and it failed mostly because the NvChad was by default installed to the `.config` directory. I use a custom dir for my `XDG_CONFIG_HOME` so I figured if anyone has the same issue this would be an easy way to have it not work for them, the fallback dir here is `.config` if the user has not set a custom one.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -98,7 +98,7 @@ _setup_terminal_shell() {
 }
 _setup_arguments() {
   # default variables to be used
-  _CONFIG_PATH="${HOME}/.config/nvim"
+  _CONFIG_PATH="${XDG_CONFIG_HOME}/nvim"
   _UPDATE=""
   _BACKUP_PATH="${_CONFIG_PATH}.bak"
   _NO_BACKUP="false"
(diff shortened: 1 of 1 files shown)


siduck76:
whats xdg config home?

Akianonymus:
> I tried installing it on my system and it failed mostly because the NvChad was by default installed to the `.config` directory. I use a custom dir for my `XDG_CONFIG_HOME` so I figured if anyone has the same issue this would be an easy way to have it not work for them, the fallback dir here is `.config` if the user has not set a custom one.

Replace the line with this

```
_CONFIG_PATH="${XDG_CONFIG_HOME:-${HOME}/.config}/nvim"
```

mmatongo:
> whats xdg config home?

It's a var that lets you pretty much keep your configuration files in a directory of your choosing and not the default `.config`

https://wiki.archlinux.org/title/XDG_Base_Directory

mmatongo:
> ```sh
> _CONFIG_PATH="${XDG_CONFIG_HOME:-${HOME}/.config}/nvim"
> ```

I replaced it with this, works now even even `$XDG_CONFIG_HOME` is unset.

siduck76 (approved):

PR #207, opened 2021-07-31
added lspsaga for testing, removed lsp config key bindings because ls…

Vanderscycle:
…psaga replaces alot of them. This configuration supports signature as you type. 
Successfully tested as I brought my config forward with NvChad.


siduck76:
@Vanderscycle you made the changes to mappings in lspconfig ( commented them ) , why tho?

siduck76:
I see some 6-7 lspsaga mappings added , what would be the equivalent of the rest of the mappings ( lspconfig default ones )

Vanderscycle:
>  @Vanderscycle you made the changes to mappings in lspconfig ( commented them ) , why tho?

More for personal preference. I realized that I never used them and lspsaga has many similar so I did override a few.

PR #209, opened 2021-08-02
Added zsh-completion plugin

ashincoder:
The following plugin will only loaded for zsh files. it is hardcoded in the database.
It doe'snt affect the startup speed


siduck76:
https://github.com/siduck76/NvChad/wiki/Chad-user-configs

siduck76:
added it in wiki https://github.com/siduck76/NvChad/wiki/Chad-user-configs#zsh-completion--compe-

PR #238, opened 2021-08-13
add unix readline's keymap in INSERT_MODE

curtainp:
when I'm coding, sometime won't exit insert mode frequently for characters move, so this unix `readline` like keymap useful.

### DIFF
--- a/lua/chadrc.lua
+++ b/lua/chadrc.lua
@@ -91,6 +91,7 @@ local M = {
             git_blame = "<leader>gb"
         },
         unix_keymap = {
+            toggle_unix_keymap = "<leader>k",
             forward = "<C-f>",
             backward = "<C-b>",
             top_of_line = "<C-a>",
--- a/lua/chadrc.lua
+++ b/lua/chadrc.lua
@@ -86,6 +86,12 @@ local M = {
             diffget_3 = "<leader>gl",
             git_blame = "<leader>gb"
         },
+        unix_keymap = {
+            forward = "<C-f>",
+            backward = "<C-b>",
+            top_of_line = "<C-a>",
+            end_of_line = "<C-e>"
+        },
         misc = {
             openTerm_right = "<C-l>",
             openTerm_bottom = "<C-x>",
--- a/lua/mappings.lua
+++ b/lua/mappings.lua
@@ -24,6 +24,12 @@ map("v", "x", [=[ "_x ]=], opt)
  this line too ]]
 --
 
+-- unix readline's keymap
+map("i", user_map.unix_keymap.forward, '<Right>', opt)
+map("i", user_map.unix_keymap.backward, '<Left>', opt)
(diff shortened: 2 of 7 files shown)


JustSage:
Personally I think this is a really bad habit and I advise against integrating this with insert mode. 
However, this could be useful in command mode.

siduck76:
@curtainp could you add two mappings for moving the cursor <up><down>?

siduck76:
> Personally I think this is a really bad habit and I advise against integrating this with insert mode.
> However, this could be useful in command mode.

might be very helpful for many so I'll add an option to enable /disable this.

lua/mappings.lua
```diff
@@ -24,6 +24,12 @@ map("v", "x", [=[ "_x ]=], opt)
  this line too ]]
 --
 
+-- unix readline's keymap
```
G-Rowell, line 27: Maybe wrap this in a lua function and check if:
- if the `chadrc.lua` var does not exist, assume this functionality should be disabled (return, early exit the function)
- if the `chadrc.lua` var exists and is true, then run this block

lua/chadrc.lua
```diff
@@ -90,6 +90,12 @@ local M = {
             diffget_3 = "<leader>gl",
             git_blame = "<leader>gb"
         },
+        unix_keymap = {
```
G-Rowell, line 93: Function mentioned in my comment below should check the table in `line 9 (above)`, `options`

G-Rowell (changes requested):
Add enable option, assume disable if option is false or does not exist

G-Rowell:
@curtainp 
Lemme know if you want some help or anything!

G-Rowell (changes requested):
This commit should be removed from your branch,

Run `git rebase --interactive HEAD~3`

change `pick` to the delete option for this commit/merge

Then instead of doing the merge to update your branch,
do `git pull --rebase upstream main` (assuming you set siduck76/NvChad as upstream)

This way you avoid adding a merge commit just to update your local branch!

Hope this helps! 

(note, because you are modifying the history of your branch, you'll have to update the online/remote copy with `git push origin --force` once your done)

G-Rowell:
Let me know if you have any questions at all,
happy to help!

curtainp:
@G-Rowell thanks, I have daily somethings bother, the fix I have finished, now to push

G-Rowell:
> @G-Rowell thanks, I have daily things bother, the fix I have finished, now to push 

No problem,
We all have other things we need to do.

@ me when you have pushed your changes or have a question

G-Rowell:
Reading now

curtainp:
> Reading now

there seems somethings wrong, i checkout a new branch to do this, wait me @G-Rowell

G-Rowell:
@curtainp 

Yeah, think you've got some git config issues
Happy to walk you through how setup a PR nicely

PR #240, opened 2021-08-14
add some useful keymap in insert/command mode with switch to toggle

curtainp:
@G-Rowell the new pr

### DIFF
--- a/lua/chadrc.lua
+++ b/lua/chadrc.lua
@@ -95,6 +95,13 @@ local M = {
             toggle_right = "<leader>v",
             toggle_bot = "<leader>h",
         },
+        unix_keymap = {
+          toggle_unix_keymap = "<leader>k",
+          forward = "<C-f>",
+          backward = "<C-b>",
+          top_of_line = "<C-a>",
+          end_of_line = "<C-e>"
+        },
         misc = {
             esc_Termmode = "jk",
             copywhole_file = "<C-a>",
--- a/lua/mappings.lua
+++ b/lua/mappings.lua
@@ -24,6 +24,45 @@ map("v", "x", [=[ "_x ]=], opt)
  this line too ]]
 --
 
+-- toggle unix readline's keymap
+map("n", user_map.unix_keymap.toggle_unix_keymap,"<cmd>lua require 'mappings'.unix_keymap()<CR>",{nowait = true})
+
+local _cmap_containp = function (key)
+  local cmap_tab = vim.api.nvim_get_keymap("c")
+
+  for _, value in ipairs(cmap_tab) do
+    if value['lhs'] == key then
+      return true
+    end
+  end
+  return false
+end
+
+M.unix_keymap = function()
+  local m = user_map.unix_keymap
+
+  if _cmap_containp("<C-A>") then
+    vim.api.nvim_del_keymap("i", m.forward)
+    vim.api.nvim_del_keymap("i", m.backward)
+    vim.api.nvim_del_keymap("i", m.top_of_line)
+    vim.api.nvim_del_keymap("i", m.end_of_line)
+    vim.api.nvim_del_keymap("c", m.forward)
+    vim.api.nvim_del_keymap("c", m.backward)
+    vim.api.nvim_del_keymap("c", m.top_of_line)
+    vim.api.nvim_del_keymap("c", m.end_of_line)
+  else
+    map("i", m.forward, '<Right>', opt)
+    map("i", m.backward, '<Left>', opt)
+    map("i", m.top_of_line, '<ESC>^i', opt)
+    map("i", m.end_of_line, '<End>', opt)
+    vim.api.nvim_set_keymap("c", m.forward, '<Right>', {noremap = true})
+    vim.api.nvim_set_keymap("c", m.backward, '<Left>', {noremap = true})
+    vim.api.nvim_set_keymap("c", m.top_of_line, '<Home>', {noremap = true})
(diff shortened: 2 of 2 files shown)


siduck76:
dang the config has gotten bigger and more complex than previous pr :(

G-Rowell:
I'm writing a review rn

lua/mappings.lua
```diff
@@ -24,6 +24,45 @@ map("v", "x", [=[ "_x ]=], opt)
  this line too ]]
 --
 
+-- toggle unix readline's keymap
+map("n", user_map.unix_keymap.toggle_unix_keymap,"<cmd>lua require 'mappings'.unix_keymap()<CR>",{nowait = true})
```
G-Rowell, line 28: Hmmm, I think rather than a mapping to enable & disable these mappings, maybe a `chadrc.lua` option 

That way a user can set the option in `chadrc.lua`

lua/mappings.lua
```diff
@@ -24,6 +24,45 @@ map("v", "x", [=[ "_x ]=], opt)
...
+      return true
+    end
+  end
+  return false
+end
+
+M.unix_keymap = function()
+  local m = user_map.unix_keymap
+
+  if _cmap_containp("<C-A>") then
```
G-Rowell, line 44: Don't need to delete keymaps if you only set them like before

G-Rowell, line 44: Yes to if, but change the conditional to checking if the option in `chadrc.lua` `options`

lua/chadrc.lua
```diff
@@ -95,6 +95,13 @@ local M = {
             toggle_right = "<leader>v",
             toggle_bot = "<leader>h",
         },
+        unix_keymap = {
```
G-Rowell, line 98: Think it could be worth moving these to the `misc` table?

As `unix_keymap` isn't a plugin, and every other table above is a plugin, whereas `misc` has some non-plugin binds?

Or even better, a sub-table in `misc`?

G-Rowell (changes requested):
Change structure to check a flag/option in `chadrc.lua`, 

instead of a mapping to enable/disable these mappings

G-Rowell:
@curtainp 

So the general gist is, this is great work but I think it should be restructured so it's a simple `if` to check a `chadrc.lua` flag,
instead of a mapping to enable/disable the mappings

curtainp:
@G-Rowell ok, thanks your suggestion, I will improve it

G-Rowell:
> @G-Rowell ok, thanks your suggestion, I will improve it

Happy to answer any questions!
If I'm not clear enough/confusing then ask & I'll clarify what I mean

G-Rowell:
> @G-Rowell ok, thanks your suggestion, I will improve it

Also, before you do your changes, modify your commit by doing `git rebase --interactive HEAD~1`
and change the commit to the edit option,

that's probs the best way to modify your PR,

(you'll need to follow the prompts, then `git push origin --force`

curtainp:
> > @G-Rowell ok, thanks your suggestion, I will improve it
> 
> Also, before you do your changes, modify your commit by doing `git rebase --interactive HEAD~1`
> and change the commit to the edit option,
> 
> that's probs the best way to modify your PR,
> 
> (you'll need to follow the prompts, then `git push origin --force`

yeah, i will be try, and report if i encounter some problems.

siduck76:
@curtainp if you dont mind , I can merge this PR and improve the features after it

G-Rowell:
Ask any questions you have at all!

@siduck76 
I'm happy to guide @curtainp through this so they learn!!:)

siduck76:
> Ask any questions you have at all!
> 
> @siduck76
> I'm happy to guide @curtainp through this so they learn!!

:chad:

curtainp:
> Change structure to check a flag/option in `chadrc.lua`,
> 
> instead of a mapping to enable/disable these mappings

@G-Rowell not get the means I think 😌

G-Rowell:
> > Change structure to check a flag/option in `chadrc.lua`,
> > instead of a mapping to enable/disable these mappings
> 
> @G-Rowell not get the means I think relieved

Okay so

G-Rowell:
Currently you enable & disable these mappings with another [mapping](https://github.com/siduck76/NvChad/pull/240/files#diff-a0766e1a2e02b2ea4daca632884f3b3da5484029148d88b3012ba2b14efcbef6R28)

instead we want to enable & disable these mappings through a flag in `chadrc.lua` [somewhere here](https://github.com/siduck76/NvChad/blob/main/lua/chadrc.lua#L9)

this area should be [an if, not a function](https://github.com/siduck76/NvChad/pull/240/files#diff-a0766e1a2e02b2ea4daca632884f3b3da5484029148d88b3012ba2b14efcbef6R41)

G-Rowell:
IE. no function, just an if which check the option you create in the `chadrc.lua` table

curtainp:
> somewhere here

ok,understand😁

G-Rowell:
Ask again if you need to!

G-Rowell:
How are you going? want any help?

siduck76:
> How are you going? want any help?

I can help too :rofl:

siduck76:
I want to update + improve this PR ASAP before having lunch and since @curtainp 's seems confused regarding the improvements ,I will merge this and improve it in the next commit :rocket:

curtainp:
> I want to update + improve this PR ASAP before having lunch and since @curtainp 's seems confused regarding the improvements ,I will merge this and improve it in the next commit 🚀

wooo, yesterday is the Chinese Saint Valentine's Day, hhhh😜

PR #251, opened 2021-08-14
Now have 3 separate terms, window, vert & hori, renamed some variables

G-Rowell:
Hi all,

Currently this creates 3 persistent terminals for you to use,
you can do this all easily with native [nvim-toggleterm.lua](https://github.com/akinsho/nvim-toggleterm.lua) functionality, but this is just a wrapper for your 3 `main` terms.

## keybinds are as follows:
- `<leader> + w` for a window term (IE. full screen)
- `<leader> + h` for a horizontal term
- `<leader> + v` for a vertical term
These are each three distinct & persistent terminals
- `JK` (not `jk`) to escape mode & hide the term
- `jk` still exits terminal mode (not changed, just noting this here)

## improvements:
- currently this saves each of the three terms as global vars, we'll likely still have to do this....but we could format it in a nicer way
-- maybe have a global nvChad table, which can contain the theme, these 3 terms...etc.

## bugs:
- I think it may be possible to `break` the system & open a term twice, will investigate this (hence this is not ready for merging)

### DIFF
--- a/lua/chadrc.lua
+++ b/lua/chadrc.lua
@@ -101,8 +101,9 @@ local M = {
         },
         toggleterm = {
             toggle_window = "<leader>w",
-            toggle_right = "<leader>v",
-            toggle_bot = "<leader>h"
+            toggle_vert = "<leader>v",
+            toggle_hori = "<leader>h",
+            hide_term = "JK"
         },
         insert_nav = {
             forward = "<C-l>",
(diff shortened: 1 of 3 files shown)


G-Rowell:
@siduck76 

Worth testing this before it's merged just to be sure

G-Rowell:
https://user-images.githubusercontent.com/3367398/129448206-6e609ca5-1978-4f6b-8d3a-1a37c57e4eee.mp4

siduck76:
>  nvChad-demo-2021-08-14_23.42.44.mp4

we have this feature already? dont we :c

PR #267, opened 2021-08-17
Separate user config from default config.

Pocco81:
This PR provides functionality for deep extending the default config with the user's config.


siduck76 (approved):

Pocco81:
Closing because it was being merged to dev and thus conflicts would've occurred with `main`, which for some reason is ahead of `dev`.

PR #279, opened 2021-08-18
added relativenumber option & statusline option

G-Rowell:
Hi all,

This is a small PR to add:
- the `relativenumber` option to `chadrc.lua` & `default_options.lua`, default is `false`
-- an autocommand to enable hybrid (number + relative) numbers in normal mode, and absolute (number) in insert mode, default is comment out
- imrpoved/fixed the statusline option to work with `NvimTree` (works with terms now too, though not enabled by default)

G


xTerma (commented):
🤨

PR #281, opened 2021-08-18
make cheatsheet enabled by default

G-Rowell:
Hi all,

As we are getting a lot of new users, I think we should enable the cheatsheet by default.

I personally use it all the time myself, and I think it's a must have IMO

G

### DIFF
--- a/lua/chadrc.lua
+++ b/lua/chadrc.lua
@@ -62,7 +62,7 @@ M.plugin_status = {
    nvim_comment = true,
    neoscroll_nvim = true,
    telescope_media = true,
-   cheatsheet = false,
+   cheatsheet = true,
 }
 
 -- make sure you dont use same keys twice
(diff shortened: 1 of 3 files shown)


siduck76:
No :l . It has some vimscript stuff in it and slows down by a bit on startup since it isnt loaded at an event , loaded after Telescope.

G-Rowell:
Using the Packer option `event = VimEnter`, we barely impact the startuptime.

Before (disabled & removed from my computer completely),
`221.150  000.005: --- NVIM STARTED ---`
`228.560  000.007: --- NVIM STARTED ---`
`220.756  000.006: --- NVIM STARTED ---`

After,
`225.747  000.006: --- NVIM STARTED ---`
`225.831  000.005: --- NVIM STARTED ---`
`228.950  000.009: --- NVIM STARTED ---`

PR #292, opened 2021-08-19
 NvChad Updater | Chadrc Fixes | Cleanup and Formatting | Misc

Akianonymus:
---

NvChadUpdater: Make update repo and url configurable | Improvr logging

---

Add NvChadUpdate command and shortcut for it

map leader+uu to it

summary of what it does:

first ask the user for confirmation and tell that the updater is gonna run git reset --hard in config repo and chadrc will be restored

take backup of chadrc in a lua string and locally in a file with chadrc.bak.(random numbers)

git reset on config dir and git pull

whether success or error, restore the chadrc file

if restore fails, then print backup file path

for more deep understanding, read the comments in utils.lua

---

feat: Do not depend on user config | Fix merging of configs

because it is a user config, so our config shoudn't break even we if dont have it

use our own table merge function

move loading config to a function

use a global variable to store the config, so no need to call the table function everytime

---

mappings|init: Move init mappings to a function, only call when required | Show err message for init

because mappings.lua is called from multiple places, so there should't be any code that executes without calling a specific function

show error message when something fails in init.lua
makes no sense to not

rearrange plugin functions alphabetically, but keep misc at top

---

tree-wide: Format files

ugh why do people don't push formatted stuff

---

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -4,5 +4,9 @@ local chad_modules = {
 }
 
 for i = 1, #chad_modules, 1 do
-   pcall(require, chad_modules[i])
+   if not pcall(require, chad_modules[i]) then
+      error("Error loading " .. chad_modules[i] .. "\n")
+   end
 end
+
+require("mappings").misc()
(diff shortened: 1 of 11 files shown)


Akianonymus:
Do not use rebase and squash, use rebase and merge.

G-Rowell:
POG

Great job Aki!!!!!

PR #306, opened 2021-08-20
Improve statusline config & hotifx close_buffer issue

G-Rowell:
Hi all,

## hotfix close_buffer issue
- the `close_buffer` lua function would exit if the current buffer was unlisted, this prevented closing buffers like `help buffers`
- this has been `hotfixed` IE. revisit this later
- initial testing makes it seem reasonable, but it is not thoroughly tested,
- it's still a big improvement on a breaking change I made (sorry....), so I think this should be pulled in
---

# removed the below commit
## fixing a design issue with statusline
- previously there was no way for the user to enable the statusline on terminals, as it was enabled in `default_config`
- this was fixed by a restructure of the `hide_statusline` table to use boolean flags,
- set an entry like so
  - `terminal = true,`  (filetype = true) true is hide, false is show
---

G

### DIFF
--- a/lua/utils.lua
+++ b/lua/utils.lua
@@ -82,6 +82,7 @@ M.close_buffer = function(bufexpr, force)
 
    local buf = vim.fn.bufnr()
    if vim.fn.buflisted(buf) == 0 then -- exit if buffer number is invalid
+      vim.cmd "close"
       return
    end
 


siduck76:
@G-Rowell this feels un-needed 

image

G-Rowell:
@siduck76

siduck76:
just put it as "nvimtree" and "terminal" , the hidden_statusline var itself feels very explicit.

G-Rowell:
@siduck76 

1. Don't merge yet please, about to update
2. What do you mean by that? We need a table in case the user wants to hide the statusline on extra filetypes?

G-Rowell:
I tested the `improve statusline` commit with a filetype of `1234-ten`, this caused no errors or bugs.
The only known issue is that the user cannot set such a filetype in the `hidden_statusline` table.

So I will improve this soon,
but this current version does not cause any bugs or errors AFAIK.

image

siduck76:
> @siduck76
> 
> 1. Don't merge yet please, about to update
> 2. What do you mean by that? We need a table in case the user wants to hide the statusline on extra filetypes?

yes , boolean values for those feels un-needed imo. just add strings of filetypes to disable statusline on those filetypes or let the user remove those strings to enable statusline in those filetypes. Like before :c

siduck76:
oh ok then ,lemme know when to merge this.

G-Rowell:
> > @siduck76
> > 
> > 1. Don't merge yet please, about to update
> > 2. What do you mean by that? We need a table in case the user wants to hide the statusline on extra filetypes?
> 
> yes , boolean values for those feels un-needed imo. just add strings of filetypes to disable statusline on those filetypes or let the user remove those strings to enable statusline in those filetypes. Like before :c

I did have it as a table of strings before,
but due to the `merging of the tables in default_config & chadrc this did not work`.

I did think about it & try that before

Hope this okay, 
should be good to merge & revisit next week

G-Rowell:
Updated initial PR comment

G-Rowell:
> I tested the `improve statusline` commit with a filetype of `1234-ten`, this caused no errors or bugs.
> The only known issue is that the user cannot set such a filetype in the `hidden_statusline` table.
> 
> So I will improve this soon,
> but this current version does not cause any bugs or errors AFAIK.
> 
> In the below image, I had just run `setlocal filetype='1234-ten`, then `echo &ft`, the echo-ed filetype is what you see at the bottom
> 
> I also tried swapping in & out of the buffer and no error was caused by the autocommand


Updated this comment with extra info

G-Rowell:
> oh ok then ,lemme know when to merge this.

I think this is ready,

need to revisit these commits later but they serve as a good fix for now

Akianonymus:
@siduck76 merge

G-Rowell:
> @siduck76 merge

+1

@Akianonymus or I will improve the statusline hide functionality when we have time.

siduck76:
> @siduck76 merge

ok!

PR #350, opened 2021-08-23
fix: don't hide dot files by default

dunglas:
Hi,

Thanks for this great config!

Hiding dot files by default is counterintuitive and hurts the developer experience as most projects relies on dot files for CI and CD (`.github/`, `.gitlab-ci.yml`...).

This PR changes the default config to show dot files.


siduck76:
@dunglas this is gonna show dotfiles by default which feels messy . 

with dotfiles : 
image

without it : 

image

Akianonymus:
I will add a option in config to customize

dunglas:
I know I can switch, but hiding dot files by default doesn't fit the IDE use case, does it?

AFAIK most IDEs show dot files by default.

Akianonymus:
@dunglas Yep, it is annoying, atleast for me.

siduck76:
> I know I can switch, but hiding dot files by default doesn't fit the IDE use case, does it?
> 
> AFAIK most IDEs show dot files by default.

oh :(

siduck76:
I'll think about this

PR #357, opened 2021-08-23
commands to load fugitive

### LINKED ISSUE #356: Lazy loading plugins by command invocation is prone to breakage/confusion
**Describe the bug**
I frequently invoke `:Gdiffsplit` from fugitive, but it doesn't seem to be available with the nvChad configs ("E492: Not an editor command: Gdiffsplit"). I'm not sure how to add a keybinding for it either, but I usually like to type in the command.

**To Reproduce**
1. Open a file in neovim inside a git repo
2. Run `:Gdiff`
3. See error

**Expected behavior**
Plugin should open a vertical split with the last commit (or optionally a specified commit/branch)

**Desktop (please complete the following information):**
 - Ubuntu 20.04
 - Alacritty
 - Neovim 0.5.0
 - NvChad 1.0 as well as 09f6c4c

21eleven:
closes #356 

Adds some fugitive commands make sense to load for. Commands for viewing diffs and for writing. Most commonly used functionality is under `Git` (ie `:Git commit`)

### DIFF
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -314,6 +314,11 @@ return packer.startup(function()
       disable = not plugin_status.vim_fugitive,
       cmd = {
          "Git",
+         "Gdiff",
+         "Gdiffsplit",
+         "Gvdiffsplit",
+         "Gwrite",
+         "Gw",
       },
       setup = function()
          require("core.mappings").vim_fugitive()


lua/plugins/init.lua
```diff
@@ -314,6 +314,11 @@ return packer.startup(function()
       disable = not plugin_status.vim_fugitive,
       cmd = {
          "Git",
+         "Gdiff",
```
ccope, line 317: Gdiff is short for Gdiffsplit. I only added Gdiffsplit here and Gdiff works for me.

PR #358, opened 2021-08-24
Add LSP code action mappings

ccope:
Tested with rust-analyzer


ccope:
I also think the mapping for vim.lsp.buf.range_formatting is wrong? (should use a `'v'` for visual selection rather than `'n'`)

siduck76:
@ccope what does code action mean?

ccope:
They are various refactoring features that are available depending on context. Examples here: https://rust-analyzer.github.io/manual.html#assists-code-actions

Issue #351, opened 2021-08-23
statusline error when performing an interactive search and replace


chebro:
**Describe the bug**
performing an interactive search and replace operation throws statuline errors

**To Reproduce**

1. open some file.
2. enter command mode and run `:%s/old/new/gc`

it throws the following error

```
replace with the (y/n/a/q/l/^E/^Y)?                                                
/home/chaos/.config/nvim/lua/plugins/configs/statusline.lua:232: attempt to index a nil value                                                                         
E15: Invalid expression: luaeval('require("galaxyline").component_decorator')("left_round")                                                                           
/home/chaos/.config/nvim/lua/plugins/configs/statusline.lua:232: attempt to index a nil value function: builtin#18 /home/chaos/.config/nvim/lua/plugins/configs/statusline.lua:232: attempt to index a nil value                                         
E15: Invalid expression: luaeval('require("galaxyline").component_decorator')("viMode_icon")                                                                          
/home/chaos/.config/nvim/lua/plugins/configs/statusline.lua:232: attempt to index a nil value function: builtin#18 /home/chaos/.config/nvim/lua/plugins/configs/statusline.lua:232: attempt to index a nil value function: builtin#18 /home/chaos/.config/nvim/lua/plugins/configs/statusline.lua:232: attempt to index a nil value         
E15: Invalid expression: luaeval('require("galaxyline").component_decorator')("ViMode")               
```

**Expected behavior**
the command should execute successfully without the `statusline.lua` error

**Desktop (please complete the following information):**
 - arch linux
 - kitty
 - v0.6.0-dev+43-g02bf251bb

G-Rowell:
I can confirm this bug

siduck76:
yea can confirm

G-Rowell:
Need all of our org to confirm before any action is taken right?

> yea can confirm



@chebro 
I'll take a good look at this tomorrow! (it's late here)

siduck76:
@G-Rowell no , its a bug lol

G-Rowell:
Lol, was just joking
Very tired and stressed

My bad, it was a poor attempt at sarcasm lol

siduck76:
bruh no sarcasms in the issues :angry:  , get serious!

siduck76:
lul no neovim + chad  = NvChad ( not a joke)

G-Rowell:
Valid points on all sides,
let's take a moment to reflect and then come back with fresh minds?

[...]

PR #362, opened 2021-08-24
BUGFIX: fix statusline bug, handle all Vim modes

G-Rowell:
Hi all,

this PR handles all Vim modes,
fixing bug in #351 

We previously weren't handling all Vim modes, causing an error in that case.

G

### DIFF
--- a/lua/plugins/configs/statusline.lua
+++ b/lua/plugins/configs/statusline.lua
@@ -216,20 +216,30 @@ gls.right[3] = {
 }
 
 local mode_colors = {
-   [110] = { "NORMAL", colors.red },
-   [105] = { "INSERT", colors.dark_purple },
-   [99] = { "COMMAND", colors.pink },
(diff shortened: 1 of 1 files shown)


siduck76:
@G-Rowell did you test it , my friend?

G-Rowell:
> @G-Rowell did you test it , my friend?

I did give it a small test for one or two of the new added mode, (not all)

it also should not have much ability to introduce bugs, as it just [adds mode entries to this table](https://github.com/NvChad/NvChad/pull/362/files#diff-fad73bee2ecbdd6a2777a5ebc2be0fbc16694846a39f44b8c5912603b0123454R228)
And [changes the format from the byte representation of the mode to the actual mode text](https://github.com/NvChad/NvChad/pull/362/files#diff-fad73bee2ecbdd6a2777a5ebc2be0fbc16694846a39f44b8c5912603b0123454R242)

PR #364, opened 2021-08-24
More lazy load | Custom mappings in config | Misc

Akianonymus:
Expose dashboard option
---
Lazy load some stuff
---
do not lazy load dashboard, barely takes time

user defer_fn to load some plugins which doesn't need to start before the main ui

load some plugins after startup even it was suppossed to be loaded on  cmd, by this we can have command auto completion too

---
utils: load_config: Use filereadable to check if file exists
---
packerInit: Print error on clone failed
---
feat: Add copy_del and copy_cut options to config
---
feat: Support for custom user mappings | Improve map function
---


nearlySplat:
Showing the dashboard on startup is a great addition, and one that I was just about to research how to do!

siduck76:
> Showing the dashboard on startup is a great addition, and one that I was just about to research how to do!



> Showing the dashboard on startup is a great addition, and one that I was just about to research how to do!

https://github.com/NvChad/NvChad/wiki#dashboard

siduck76:
update the lazy load for gitsigns

siduck76 (approved):

PR #392, opened 2021-08-27
Changed mappings to move between windows/terminals with Ctrl+h,j,k,l

JoakoV3:
I love how these configs are getting better every day!

There was no easy movements between terminals, vimtree nor split opened buffers. With this change, you can move between all of them easily without the mouse by pressing `Ctrl + h` ; `Ctrl + l` ; `Ctrl + j` ; `Ctrl + k` . I thought it should be having it by default.

### DIFF
--- a/lua/chadrc.lua
+++ b/lua/chadrc.lua
@@ -143,6 +143,11 @@ M.mappings.plugin = {
    bufferline = {
       next_buffer = "<TAB>", -- next buffer
       prev_buffer = "<S-Tab>", -- previous buffer
+      --better window movement
+      moveLeft = "<C-h>",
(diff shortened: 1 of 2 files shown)


Esequiel378 (approved):
Love this key mapping 👌

ashincoder (approved):
:+1:

siduck76:
> ctrl + h is used for nvim tree focus

can we change it to leader nf?

ashincoder:
better move nvim-tree toggle to `<leader>e`

G-Rowell:
> better move nvim-tree toggle to `<leader>e`
> 

@siduck76

PR #412, opened 2021-08-30
Improve startup time by speeding up lua module requirements

Pocco81:
### Who is getting benefited?

+ NvChad modules that are required after the impatient plugin (the first plugin to be loaded) gets loaded.
+ Plugins that are not being lazy loaded (e.g. `NvChad/extensions`, `nvim-lua/plenary.nvim`, user plugins they decide not lazy load, etc...)

### Proof:
+ Here is a[ comparison table](https://github.com/lewis6991/impatient.nvim#performance-example) made by the author of the plugin, where it loads plugins with and without impatient.nvim.
+ Personal test: regarding plugins, I don't lazy load `nvim-cmp`: (measured using Packer.nvim's profiler)

Without impatient.nvim:
image

With impatient.nvim:
image

### DIFF
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -5,14 +5,13 @@ if not present then
 end
 
 local use = packer.use
-require('impatient')
+require "impatient"
 
 return packer.startup(function()
-
-	use {
-		"lewis6991/impatient.nvim",
-		rocks = "mpack",
-	}
+   use {
+      "lewis6991/impatient.nvim",
+      rocks = "mpack",
+   }
 
    local plugin_status = require("core.utils").load_config().plugin_status
 
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -5,8 +5,15 @@ if not present then
 end
 
 local use = packer.use
+require('impatient')
 
 return packer.startup(function()
+
+	use {
+		"lewis6991/impatient.nvim",
+		rocks = "mpack",
+	}
+
    local plugin_status = require("core.utils").load_config().plugin_status
 
    -- this is arranged on the basis of when a plugin starts


siduck76:
@Pocco81 I guess we must wait until this stuff gets merged into upstream since we would require lua51-mpack as a dependency for this.

lua/plugins/init.lua
```diff
@@ -5,8 +5,14 @@ if not present then
 end
```
Diego77777777777, line 5:
```suggestion
 end
```

Diego77777777777: Un error

Diego77777777777 (approved):

Diego77777777777:
****

Diego77777777777:
Te hice varias varios revisiones

Diego77777777777:
> #### This is being done using [impatient.nvim](https://github.com/lewis6991/impatient.nvim).
> 
> ### Who is getting benefited?
> 
> + NvChad modules that are required after the impatient plugin (the first plugin to be loaded) gets loaded.
> + Plugins that are not being lazy loaded (e.g. `NvChad/extensions`, `nvim-lua/plenary.nvim`, user plugins they decide not lazy load, etc...)
> 
> ### Proof:
> + Here is a[ comparison table](https://github.com/lewis6991/impatient.nvim#performance-example) made by the author of the plugin, where it loads plugins with and without impatient.nvim.
> + Personal test: regarding plugins, I don't lazy load `nvim-cmp`: (measured using Packer.nvim's profiler)
> 
> Without impatient.nvim:lynux
>image
> With impatient.nvim:
>image
>

Diego77777777777:
[]()

Issue #372, opened 2021-08-25
I'd like to turn off some options in the plugins 


macarthy:
**Is your feature request related to a problem? Please describe.**

the virtual text in lsp dianostics is very distracting, I would perfect a sidebar icon and an underlineding the issue , hover over the line the issue and maybe show it in the status bar

**Describe the solution you'd like**
I'd would prefer a sidebar icon and an underlined issue, that you could hover over to get info 

In general if we could create a {{filename}}.local.lua file that defines extra plugins  or mapping etc that would not get trashed on the up that would be great

I'd really like hop.nvim  and test-vim

benjamin-shih:
Hey,

I found that if you would like to just disable the virtual text, you can navigate to /lua/plugins/configs and in lspconfig.lua edit the virtual_text variable to false as such: 
```
vim.lsp.handlers["textDocument/publishDiagnostics"] = vim.lsp.with(vim.lsp.diagnostic.on_publish_diagnostics, {
virtual_text = false,
-- {
--   prefix = "",
--   spacing = 0,
-- },
signs = true,
underline = true,
update_in_insert = false, -- update diagnostics insert mode
})
```

macarthy:
Yeah that is what I did but it gets trashed on update.

siduck76:
> Yeah that is what I did but it gets trashed on update.

no worries , https://github.com/NvChad/NvChad/pull/368 has been merged to test branch and soon it'll get merged to main!

G-Rowell:
Hey @macarthy @benjamin-shih 

The PR by Morten & I, #465 will fix this!

It currently has an `override` for changing the `lsp diagnostics`, which is exactly what you seem to want!
You'll be able to `override` this from within your `lua/custom/` area, so it will be protected from conflicts hopefully!


Should we close this when the PR is merged?

G

siduck76:
the PR is merged , check the config docs https://nvchad.github.io/config

PR #414, opened 2021-08-31
feat: Hook based option overrides

morten-olsen:
Hey again.

The issue in #372 does not really get solved by the current hook system, as it only adds the ability to run additional setup during NvChad initialization/commands, not modify the existing stuff.
This is something I was originally wondering if I should add to that original PR but decided against it, as I wasn't sure it is a feature that is generally useful enough, but well given that it could solve some issues, here it is for discussion:

So this introduces hook based overrides.

Of cause `chadrc.lua` should be the main place to configure NvChad, but if all options of all plugins was exposed it would become so extensive that it would be really hard to maintain, so I get having some values be set to something that just works for most people makes sens.

The idea here is that for the rest, it could construct its options using the hook system, using `overrides.get` which would just return the object if no override hooks are registered and everything just-works™. If the user needs to modify any of these "deep" options, it can de done inside `/lua/custom` by registering an override hook, which receives the option, and can then modify it or return something completely different.

This could allow power users or users with very specific requirements access to those deep options, while keeping chadrc to just the options that a lot of people would want to tweak for their need.

I have for illustrative purposes set the specific options from #372 up for override and added an override to `custom/init.lua`, but to be truly useful, all options that power users should be able to override should be wrapped in `overrides.get`

Again, i originally wanted to include this, but decided against it as I don't know if this is the right solution, but could maybe be of use for handling some configuration cases, what do you guys think?

### DIFF
--- a/lua/core/hooks.lua
+++ b/lua/core/hooks.lua
@@ -1,4 +1,4 @@
-local hooks, M = {}, {};
+local hooks, overrides, M = {}, {}, {};
 local allowed_hooks = {
   "install_plugins",
   "setup_mappings",
@@ -34,5 +34,32 @@ M.run = function(name, args)
     hook(args)
   end
 end
+
+M.createOverrides = function(module)
+  local O = {};
+
+  O.get = function(name, default)
+    local current = default;
+    if overrides[module] and overrides[module][name] then
+      for _, override in pairs(overrides[module][name]) do
+        current = override(current)
+      end
+    end
+    return current;
+  end
+
+  return O;
+end
+
+M.override = function(module, name, fn)
+  if overrides[module] == nil then
+    overrides[module] = {};
+  end
+  if overrides[module][name] == nil then
+    overrides[module][name] = {};
+  end
+  table.insert(overrides[module][name], fn)
+end
+
 
 return M;
--- a/lua/custom/init.lua
+++ b/lua/custom/init.lua
@@ -4,6 +4,6 @@
 local hooks = require "core.hooks"
 
 hooks.override("lsp", "publish_diagnostics", function(current)
-  current.virtual_text = false;
-  return current;
+   current.virtual_text = false
+   return current
 end)
--- a/lua/custom/init.lua
+++ b/lua/custom/init.lua
@@ -1,2 +1,9 @@
 -- This is where you custom modules and plugins goes.
 -- See the wiki for a guide on how to extend NvChad
+
+local hooks = require "core.hooks"
+
+hooks.override("lsp", "publish_diagnostics", function(current)
+  current.virtual_text = false;
+  return current;
+end)
--- a/lua/plugins/configs/lspconfig.lua
+++ b/lua/plugins/configs/lspconfig.lua
@@ -1,6 +1,6 @@
 local present1, lspconfig = pcall(require, "lspconfig")
 local present2, lspinstall = pcall(require, "lspinstall")
-local overrides = require("core.hooks").createOverrides("lsp")
+local overrides = require("core.hooks").createOverrides "lsp"
 
 if not (present1 or present2) then
    return
@@ -122,7 +122,7 @@ lspSymbol("Information", "")
 lspSymbol("Hint", "")
 lspSymbol("Warning", "")
 
-local lsp_publish_diagnostics_options = overrides.get('publish_diagnostics', {
+local lsp_publish_diagnostics_options = overrides.get("publish_diagnostics", {
    virtual_text = {
       prefix = "",
       spacing = 0,
@@ -131,7 +131,10 @@ local lsp_publish_diagnostics_options = overrides.get('publish_diagnostics', {
    underline = true,
    update_in_insert = false, -- update diagnostics insert mode
 })
-vim.lsp.handlers["textDocument/publishDiagnostics"] = vim.lsp.with(vim.lsp.diagnostic.on_publish_diagnostics, lsp_publish_diagnostics_options)
+vim.lsp.handlers["textDocument/publishDiagnostics"] = vim.lsp.with(
+   vim.lsp.diagnostic.on_publish_diagnostics,
+   lsp_publish_diagnostics_options
+)
 vim.lsp.handlers["textDocument/hover"] = vim.lsp.with(vim.lsp.handlers.hover, {
    border = "single",
 })
--- a/lua/plugins/configs/lspconfig.lua
+++ b/lua/plugins/configs/lspconfig.lua
@@ -1,5 +1,6 @@
 local present1, lspconfig = pcall(require, "lspconfig")
 local present2, lspinstall = pcall(require, "lspinstall")
+local overrides = require("core.hooks").createOverrides("lsp")
 
 if not (present1 or present2) then
    return
@@ -121,7 +122,7 @@ lspSymbol("Information", "")
 lspSymbol("Hint", "")
 lspSymbol("Warning", "")
 
-vim.lsp.handlers["textDocument/publishDiagnostics"] = vim.lsp.with(vim.lsp.diagnostic.on_publish_diagnostics, {
+local lsp_publish_diagnostics_options = overrides.get('publish_diagnostics', {
    virtual_text = {
       prefix = "",
       spacing = 0,
@@ -130,6 +131,7 @@ vim.lsp.handlers["textDocument/publishDiagnostics"] = vim.lsp.with(vim.lsp.diagn
    underline = true,
    update_in_insert = false, -- update diagnostics insert mode
 })
+vim.lsp.handlers["textDocument/publishDiagnostics"] = vim.lsp.with(vim.lsp.diagnostic.on_publish_diagnostics, lsp_publish_diagnostics_options)
 vim.lsp.handlers["textDocument/hover"] = vim.lsp.with(vim.lsp.handlers.hover, {
    border = "single",
 })


morten-olsen:
Correction: that specific issue could be fixed with the "current" hook system by just re-setting it in the ready hook I guess

marwan38:
The problem with this approach is that lsp won't be able to infer the types of the `current` object. Type inference makes it much easier for the user to make configuration changes without having to go anywhere.

Take lunarvim for example, I was able to make most my modifications right away, without any guidance from anywhere.

So IMO config based would be the way to go, if the maintainers of this repo indeed want to head that way.

morten-olsen:
@marwan38 Agree that configurability is prefered but exposing all LSP configuration options in a config file doesn't seem feasible, so this is only intended to offer some level of configurability in places that wouldn't make sense to add to `chadrc` because of scope.

For the specific issue referred to in the comment, that could be solved with the upcoming hook system (without the override system) by adding

```lua
require("core.hooks").add("ready", function # Invoked after NvChad is done setting up
  vim.lsp.handlers["textDocument/publishDiagnostics"] = vim.lsp.with(vim.lsp.diagnostic.on_publish_diagnostics, {
    virtual_text = false
    underline = true,
    update_in_insert = false, -- update diagnostics insert mode
  })
end)
```

to `lua/custom/init.lua` (Which shouldn't be affected by updates). Note that it uses the objects directly, so initial type inference should work here.

G-Rowell:
I do like this idea,

I think it solves a problem that @morten-olsen describes very well, and I certainly agree that we need to either:

1. Expose every single configurable option of every plugin we use in `chadrc.lua`
or
2. Allow users to completely override the plugin setup of any reasonable plugin

Clearly I think this PR (which is approach # 2) is best!

---

Eg: there may be a plugin option that can only be set once during setup, latter modifications to this plugin option may not register, hence the need for a proper override like this PR.

I'm not sure I completely understand this issue:
> The problem with this approach is that lsp won't be able to infer the types of the current object.

(My understanding of Lua may not be as good as you two :smile: )

G-Rowell:
If I'm understanding correctly:

To allow a user to override a default config: 
- in the `lua/plugins/configs/<PLUGIN CONFIG FILE>`
   - import the override hook `local overrides = require("core.hooks").createOverrides "<PLUGIN>"`
   - where applicable, wrap the existing config like so:
   
   ```
   local config_func = overrides.get("<CONFIG OVERRIDE HOOK>, {
   < normal / default plugin config here >
   })

   plugin.config(config_func)
   ```

morten-olsen:
> I'm not sure I completely understand this issue:

Tbh I don't either, this is basically some of the first lua I've ever written, so I have made three assumptions which it is based on

- Lua actually having a type system/type inference
- vim.lsp.handlers["textDocument/publishDiagnostics"] actually have types/ can inferred type correctly
- Lua can make type match comparisons between the implicit table and the handlers type

Don't know if this is true, but given the original question I guessed that it is either true in both cases or false in both cases, so if it could be done without using the hook, it should work with hooks as well

Guess for the override hooks those object could not be type inferred in any case since the implementation has no idea what a user override might return

G-Rowell:
Hmmmmm, @morten-olsen 

I'll have to take a look in the next day or two, (it's 11pm here for me)

I'm thinking I'll test your earlier custom hooks PR and once that's approved I'll come have a good look at this?


Do you have anything in particular you want me to check or look at?


G

morten-olsen:
@G-Rowell no rush, rather right than fast :smiley: 

I personally think the initial hook proposal (The one in `test`) is a useful addition, and should work as advertised. I think the main thing here is if `chadrc.lua` should be moved in to `/custom`, so the update script can ignore everything in, and nothing but `/custom`. And if `/custom/init.lua` should be removed from the repo and just be documented instead.

In terms of this here PR, again, I'm pretty sure it will work as advertised, the question is if this is the direction that NvChad should move in. As i see it there are three options:

* Add everything to chadrc (Making it really bloated)
* Use something like this PR so people with specific need could configure
* Have non-changable options in NvChad to ensure a good common user experience, but have people with specific need fork it to make changes to deep options, with the merge issues that follows.

I honestly don't know which would be the right approach, and will let that be up to you guys to decide (Though I would prefer to avoid the bloated chadrc).

G-Rowell:
@siduck76 

I think merge this now (it's being merged into `test`),
and then we'll push them into `main` as soon as were all happy/in agreement?

@morten-olsen 
I think this is pretty awesome, I think I may wrap some of NvChad's config in override hooks, to enable greater config
Thank you again for your awesome work, will keep you updated! :grin: 


G

siduck76:
@morten-olsen if this PR is finished , shall I merge it into test?

G-Rowell:
I'm pretty certain it's finished, but happy to wait for Morten

morten-olsen:
Yeah, Think the only thing is that the addition to custom/init.lua should be remove that was only for reference

G-Rowell:
@morten-olsen 

Yeah, I think so.

I'm currently testing & working on any small finishing touches.
I'll probably make a PR and tag you in it (it'll be 90% your actual commits, so don't worry!)

Will keep you in the loop.

I agree about the `init.lua`, also thinking about changing the `chadrc.lua` situation tbh


G

G-Rowell:
@morten-olsen 

Hey Morten,

I've been working through and trying to fully understand your two PRs,
one thought I've had is to move to a different design for plugin overrides.

Instead of overriding specific parts of a given plugin config, instead override the whole config for the given plugin.

---

Currently, we would have to expose a certain block within an existing configuration file to allow a user to override it.
I feel like this would require us to officially support every edge case of customisation that any user wants.
[Example of this issue](https://github.com/NvChad/NvChad/pull/414/files#diff-9f2ea60dcf47caf192197bcff8823913b3b46441e32bb8b384d13b2a9385c67eR125)

---

Instead I propose we simply override the whole config file, this way the user can either take full control over the plugin or leave it.
This would place a more appropriate (but not rediculous) amount of work on the user, as we don't have to support every edge case of customisation.

An example would be to have a check in this [section to decide whether to call the override or run normally](https://github.com/NvChad/NvChad/blob/main/lua/plugins/init.lua#L49)

---

What are your thoughts?

G

G-Rowell:
```
      config = function()
           if hook_overwrite_exists then
             require("custom.blankline")

           else
             require("plugins.configs.others").blankline()
           end
         end,
```

G-Rowell:
Probs a nicer way, just a quick thought before I look properly tmr

abalmos:
@G-Rowell For what is is worth, as a NvChad user it would defeat the entire purpose of using it if I had to re-configure an entire plugin just to make a change or two.

Plus, this would bring back the original problem. Not getting configuration upgrades without manual intervention.

G-Rowell:
> @G-Rowell For what is is worth, as a NvChad user it would defeat the entire purpose of using it if I had to re-configure an entire plugin just to make a change or two.
> 
> Plus, this would bring back the original problem. Not getting configuration upgrades without manual intervention.

Oh for sure, I understand what you mean and appreciate the input.

The big problem is that you currently cannot modify any setting within a plugin at the moment without risking complex merge conflicts.

So it wouldn't be a perfect solution, but I don't think there is any other way imo

G-Rowell:
@abalmos 

The alternative is that every single time a user wants to modify a plugin setting within `lua/plugins/config/<plugin config file.lua` NvChad has to specifically make effort to accommodate that.


At some point the user has to take responsibility if they wish to modify NvChads bundled & pre-configured plugins.

(Though any additions should hopefully never cause issues (IE. Custom functions & added plugins on top of NvChad)

abalmos:
Isn't that the point of this PR though? Admittedly I don't yet understand enough of NvChad to grok how it is accomplishing overrides, but the example suggests it can. 

To make NvChad usble for my workflow I required two changes to plugins which I have lost a few times now forgetting that an upgrade resets everything rather then just trying a merge.

abalmos:
By the way, that's not meant to be critical, I have enjoyed NvChad. Just reenforcing that not just power users need to make changes and that's a bit difficult to manage at the moment. 

Without more context, this PR seems to solve that problem. I could be miss understanding though.

G-Rowell:
Yes, this PR is to enable overriding default plugin configs.

But there has to be code within the default plugin configs to allow such a thing.

Take a look at the modified files of the PR, the lsp config can only be overwritten in certain places, and only where we have programmed in such functionality.


Hence:
Decide what parts (or all) of a given config can be overwritten.

If not all, then inevitably users will create an issue because they cannot overwrite a given section of a given config.

---


Thus I propose users take some responsibility for such a thing and:
- build in common options (see `chadrc`)
- allow users to completely overwrite the default config for a given plugin

G-Rowell:
@abalmos 
It's late in my timezone, feel free to post a few comments and I'll take a look tmr!

Regardless, always appreciate input and suggestions! :)

G-Rowell:
> By the way, that's not meant to be critical, I have enjoyed NvChad. Just reenforcing that not just power users need to make changes and that's a bit difficult to manage at the moment. 
> 
> Without more context, this PR seems to solve that problem. I could be miss understanding though.

Totally fine! It's not taken in a negative manner at all, 
Always appreciate input.

Will be back here tmr to discuss!

morten-olsen:
@G-Rowell actually I agree with your suggestion.After thinking it through I find a few issues with my approach

* While all options could be made configurable through override hooks, even then, if the user wanted to add or remove an invocation that could not be done through this system, so it only adds partial configurability, where your suggestion adds full configurability
* While only modifying some aspects of an options object allows customization while still getting updates to plugin configuration, that is also a problematic route. If an option object changes in NvChad or gets removed that could break the users customization in unforeseeable ways.

So yeah, again think your idea is better, even though it would be cool to only partial changes to a plugins configuration while still getting updates to the rest, it could lead to a lot of frustration where users setup start breaking without them having changed anything, so more safe that a user who wants to make changes to a plugin not exposed through chadrc would have to take full control over that plugins configuration.

G-Rowell:
@morten-olsen 

Do you think there should be any changes to the override system for this new design?
I think we can leave it as it is maybe,
have an override module `configs`, which has the overrides for registering configs in `lua/plugins/init.lua`.

Then @abalmos brings a great point, so we expose a lot more options through `default_config` (thus can be changed in `chadrc`), 
this way users:
- have a lot more plugin options to change from within `chadrc` (maybe preventing their need to overwrite a whole plugin config)
- allow users to completely overwrite a plugin config using this overwrite PR


I'm happy to do most of the work here, as I am on uni break this week.
Though this is your PR, so totally happy to let you do so if you wish.

G

PR #417, opened 2021-08-31
fix(plugin): indentline using config instead of setup

marwan38:
I wanted to set up indent line `show context` and it wasn't working properly.

Your original packer definition of this plugin is using `config`. I think that defining global variables, ie `vim.g.indent_line_do_something` needs to be declared within the packers `setup` function (run before the plugin). 

I went through their docs and saw that they have a setup function instead, and that's usually better than riddling the global scope with variables.

### DIFF
--- a/lua/plugins/configs/others.lua
+++ b/lua/plugins/configs/others.lua
@@ -43,14 +43,14 @@ M.better_escape = function()
 end
 
 M.blankline = function()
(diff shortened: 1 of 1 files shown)


ashincoder (approved):
:+1:

siduck76 (approved):

PR #421, opened 2021-09-01
fix(indentline): move global vars into the indentline setup call

marwan38:
@siduck76 I missed an important part of the docs:

```
Setup
To configure indent-blankline, either run the setup function, or set variables manually. The setup function has a single table as argument, keys of the table match the :help indent-blankline-variables without the indent_blankline_ part.
```

Would you double check that its ok? I think it looks fine to me?

image

### DIFF
--- a/lua/plugins/configs/others.lua
+++ b/lua/plugins/configs/others.lua
@@ -43,13 +43,10 @@ M.better_escape = function()
 end
 
 M.blankline = function()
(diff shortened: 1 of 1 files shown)


siduck76:
oh lul , I saw your message on discord regarding this indentline config and made the change asap . Looked at the PRs just now xD. Ig the changes have been made already so ima close this 

https://github.com/NvChad/NvChad/commit/5742567630009a2f88a74614b3ff2330a91525e1

PR #443, opened 2021-09-07
add jellybeans theme

boppyt:


boppyt:
image

boppyt:
swapped `cyan` and `blue`

siduck76:
@boppyt you gotta check which color does so the theme colors look perfect 

image


:(

PR #457, opened 2021-09-10
Create M.plugins section

azndd:
https://nvchad.netlify.app/getting-started/post-install#install-lsp-servers points to M.plugins, even though this section doesn't exist.


siduck76:
It exists https://github.com/azndd/NvChad/blob/c374d59f7a124f19d83afcb46cf025b9a78026eb/lua/chadrc.lua#L221

azndd:
Yeah, forgot that chadrc is left untouched on an update.

PR #298, opened 2021-08-19
feat: Added highlight on yank.

ashincoder:

PR #461, opened 2021-09-12
Improved mappings

CT3:
1.  Added zz (align)  to some search jumps and lsp jumps
2. J without moving the cursor
3. Added some undo breaks  so when clicking u it will not undo full sentences 
4. Added handy way of dragging lines up and down
5.Added highlight on yank (so you can see what is being copied)


bruhtus:
You might want to edit the comment, there's some type there. `allign`, `midle`, `withou`, `undoo`, `these m`.

siduck76:
Highlight on yank feels flashy so we removed it before .

CT3:
Highlight on yank is nice for a new users like me, but I see how it would get annoying for experienced users who feel like they know what they are doing.

siduck76:
> Highlight on yank is nice for a new users like me, but I see how it would get annoying for experienced users who feel like they know what they are doing.

can you show how does it look

siduck76:
https://github.com/NvChad/NvChad/pull/298#issuecomment-902067183 it flashes for a second :(

CT3:
yes it highlights text you yanked for a second.

lua/core/mappings.lua
```diff
@@ -27,6 +27,31 @@ M.misc = function()
 
       -- use ESC to turn off search highlighting
       map("n", "<Esc>", ":noh <CR>")
+       ---- jump and align in the middle
```
G-Rowell, line 30: Can you use https://nvchad.netlify.app/contribute#how-to-remove-or-edit-commits-from-your-pr

To remove the `fix typos commit`, using `squash` please! :)

lua/core/mappings.lua
```diff
@@ -43,16 +43,6 @@ M.misc = function()
       map("n", "<leader>j", ":m .+1<CR>==", opt)
       map("v", "J", ":m '>+1<CR>gv=gv", opt)
       map("v", "K", ":m '<-2<CR>gv=gv", opt)
-      map("n", "<leader>lf", ":Lspsaga lsp_finder<CR>", opt)
```
G-Rowell: Hmmm, maybe I've missed it, but why is this PR removing these mappings?

Think `lsp-saga` has been removed from `NvChad:main`, so consider dropping this commit?

(use the same link to our docs as , with `drop`)

lua/plugins/configs/lspconfig.lua
```diff
@@ -32,8 +32,8 @@ local function on_attach(_, bufnr)
    buf_set_keymap("n", "<space>ca", "<cmd>lua vim.lsp.buf.code_action()<CR>", opts)
    buf_set_keymap("n", "gr", "<cmd>lua vim.lsp.buf.references()<CR>", opts)
    buf_set_keymap("n", "<space>e", "<cmd>lua vim.lsp.diagnostic.show_line_diagnostics()<CR>", opts)
-   buf_set_keymap("n", "[d", "<cmd>lua vim.lsp.diagnostic.goto_prev()<CR>", opts)
-   buf_set_keymap("n", "]d", "<cmd>lua vim.lsp.diagnostic.goto_next()<CR>", opts)
+   buf_set_keymap("n", "[d", "<cmd>lua vim.lsp.diagnostic.goto_prev()<CR>zz", opts)
```
G-Rowell, line 35: This seems like a good & reasonable idea!! 😄

G-Rowell (changes requested):
Like it overall, small cleanup to do maybe?

Correct me if I'm wrong about anything 😄 


Thanks for the PR!
G

lua/plugins/configs/lspconfig.lua
```diff
@@ -32,8 +32,8 @@ local function on_attach(_, bufnr)
...
+   buf_set_keymap("n", "]d", "<cmd>lua vim.lsp.diagnostic.goto_next()<CR>zz", opts)
```
siduck76, line 36: @G-Rowell I think most of the part in this PR ( except for this lsp keymap) is user specific , can be added after the custom hooks PR is merged right?

ashincoder: I don't think these are user specific. They should be builiting. Cause these are diagnostic prevs

G-Rowell: Agree with ashin, these could do well in base NvChad

siduck76: > I don't think these are user specific. They should be builitin. Cause these are diagnostic prevs

No i dont mean them , I meant these : 

image

ashincoder: bruh these are improvements in the builiting vim bindings not user bindings.

G-Rowell: Hmmmm, 

I do think that some of these mappings should be checked/tested.
Also they hard code their mapping key, rather than grab from `default_config` / `chadrc`.

I think we've also tried, and not been keen on `highlight on yank` being included by default?


Hmmm, not sure

siduck76: > bruh these are improvements in the builiting vim bindings not user bindings.

bruh the user can add this himself once the custom hooks PR gets merged

ashincoder: > Hmmmm,
> 
> I do think that some of these mappings should be checked/tested.
> Also they hard code their mapping key, rather than grab from `default_config` / `chadrc`.
> 
> I think we've also tried, and not been keen on `highlight on yank` being included by default?
> 
> Hmmm, not sure

check primes video on must add keymappings for vim. https://www.youtube.com/watch?v=hSHATqh8svM

ashincoder: > Hmmmm,
> 
> I do think that some of these mappings should be checked/tested.
> Also they hard code their mapping key, rather than grab from `default_config` / `chadrc`.
> 
> I think we've also tried, and not been keen on `highlight on yank` being included by default?
> 
> Hmmm, not sure

Maybe highlight on yank may not be needed and we could add it as an option for the user or smtng

siduck76: Dont think so , once the custom hooks PR is merged , that can be just added as an autocmd

ashincoder: Have u seen primes video. These are must add mappings. The highlightonyank might not be needed but others are

siduck76: i dont watch prime's vids :angry: :smile:

ashincoder: I watch em.

G-Rowell: Would agree with Ashin here,
though this PR does need some small housekeeping regardless

bruhtus: > bruh the user can add this himself once the custom hooks PR gets merged

I agree with siduck here, why not make a dedicated section for `recommended default remaps` in wiki or something rather than hardcoded the default remaps like this? I think it would be easier for new user to add new mapping rather than remove existing mapping.

G-Rowell: While I do think that's a fair point,

NvChad is already pretty opinionated out of the box, that's what it's designed to be.

We want to provide great & reasonable defaults to everyone, so the question becomes:

> Is this mapping a good & sensible default, is it non-language specific & something that a majority of users would beneift from?

I think the answer is yes here imo.

Though I completely agree that we should always aim to have some way of turning off such things

bruhtus: Adding an option to turn off the default would be ideal. I just don't want the tragedy of vim-sneak happen here too. Imagine you need to make a mapping just to disable the default mapping like this `nnoremap <F99> <plug>mapping1<plug>mapping2` (that's actually what justinmk suggested). Ugh, that's horrible.

siduck76: I agree with @bruhtus , nvchad needs to have less config as possible since its meant to be used as a base config and the user adds his/ her stuff on the top of it.

G-Rowell: Oh lord yeah, using a mapping to disable a plugin is not ideal.

We already have `default_config` / `chadrc` options to disable some in-built features.

[Like here](https://github.com/NvChad/NvChad/blob/main/lua/default_config.lua#L54] (though this should be put in a separate table to normal NeoVim options)

G-Rowell:
Hi @CT3 @MantasPrevayl 

Thanks for the PR!

would you possible consider renaming your commit with something akin to:
https://www.conventionalcommits.org/en/v1.0.0/
```
docs: correct spelling of README
```

Would also prefer if you remove the `Merge branch....` `Merge remote-tracking branch....` commits.

An easy way to prevent these messages is
```
git pull --rebase
```



Sorry for being so pedantic!
Really appreciate the PR as always! 😁

G

lua/plugins/configs/lspconfig.lua
siduck76: what does the zz mean here?

G-Rowell: `zz` will try to focus your current line,

IE. it will try to scroll to position your current line in the center of the screen

siduck76: @G-Rowell all of the PR is mostly user specific! and ig it'd be better to have lspconfig's default mappings 

image


If the user doesnt want it then the overriding default config feature could be used , since #465 has merged ._.

PR #465, opened 2021-09-14
WIP + BREAKING - Custom config overhauls

G-Rowell:
Heyo all,

This PR is a big overhaul of the custom config side of NvChad.

Huge work in this has been done by @morten-olsen 

Hopefully this signals the next step in NvChad being a complete base for users, who can then customize it as they wish!

---

## WIP Work in progress

**Please do not merge this yet**

---

## Breaking:

Users will not be able to update with the `NvChad updater`, instead **use git to update directly**,
this will **bring merge conflicts, but will hopefully be some of the last ever merge conflicts ever for NvChad!!**

Probably best to do:
```
cp lua/chadrc.lua lua/chadrc.lua.BAK
git pull --rebase
```
(assuming you've just cloned the `NvChad/NvChad` repo directly without forks/upstream remotes)

---

## Todo:
- [x] add more documentation to `default_config` to explain it
- [ ] add more `override`s to various plugin tables
- [ ] add nvchad website docs & examples
- [ ] change `NvChad/extensions` to new `default_config` & `chadrc` path, I have a commit ready to push to do so
- [x] change to conventional commits https://www.conventionalcommits.org/en/v1.0.0/

---

## Changes:
- Allow complete override of default/packaged plugin configs from `chadrc`
- Greatly cleaned up `chadrc`
   - (it was a duplicate of `default_config`, but it only needs the users custom changes in it, so it should be just a blank template)
- MORTEN: added `hooks functionality`, to allow actions to run on/do:
   - `install_plugins`
   - `setup_mappings`
   - `ready` (IE. the last config actions on boot)
- MORTEN: added `override functionality`, to allow overriding a table within a given plugin config file
   - **NOTE: we have to manually add the override tool where appropriate, I am working on this now**
   - This could be expanded in the future to allow overriding functions as well (I did a lot of work experimenting with this myself)
- Moved `chadrc -> custom/chadrc`, `default_config -> core/default_config`
- Cleaned `default_config`

---

Credit must go to Morten Olsen, for his fantastic ground work on this PR!

Thanks,
G


lua/core/default_config.lua
```diff
@@ -66,8 +65,10 @@ M.options = {
    ruler = false,
    updatetime = 250,
    -- used for updater
-   update_url = "https://github.com/NvChad/NvChad",
```
siduck76, line 69: this made more sense imo , nv_chad doesnt :p

lua/core/default_config.lua
```diff
@@ -148,7 +152,7 @@ M.mappings.plugin = {
       next_buffer = "<TAB>", -- next buffer
       prev_buffer = "<S-Tab>", -- previous buffer
    },
-   chadsheet = {
```
siduck76, line 151: nooooooooooo :sob: :laughing:

lua/core/default_config.lua
```diff
@@ -160,7 +164,7 @@ M.mappings.plugin = {
       new_file = "<leader>fn", -- basically create a new buffer
       open = "<leader>db", -- open dashboard
       session_load = "<leader>l", -- load a saved session
-      session_save = "<leader>s",
+      session_save = "<leader>s", -- save a session
```
siduck76, line 167: no need to add a comment there , also remove the above comment too .

session_save/ load itself makes sense .

G-Rowell: Hmmmmm, I'm okay with being a little verbose with comments & explanations here, but if you really think I'll happily remove it!

lua/core/default_config.lua
G-Rowell: Yes :chad :

It's confusing, I think it's funny, but too confusing

Soz dude

lua/core/default_config.lua
G-Rowell: Fair, happy to change it back,

I do think having a separate section for internal NvChad options could be good, but happy to revert this

lua/core/default_config.lua
ashincoder: Being a lil verbose doesn't harm at all

siduck76: > Hmmmmm, I'm okay with being a little verbose with comments & explanations here, but if you really think I'll happily remove it!

yes remove the comments , the var names are already very self-explanatory

ashincoder: M. But it doesn't harm adding more verbose comments. It is used by the user right so .

lua/core/default_config.lua
G-Rowell: Changed back, can always revisit later if we need to

lua/core/default_config.lua
G-Rowell: It's `default_config` so we can always be verbose, but I'll leave this open for now as I should re-do all the comments in that file anyway

vladdoster:
I am glad I'm not the only one to change `chadsheet` to `cheatsheet` :)

---

My two cents about a trend I am seeing across nvim configs...

Let me preface by stating this isn't related to your PR or changes. Kudos to you for the work in this PR.

These changes are being made with good intentions, but a few issues I see...

1. The changes seem unnecessary and approaching over engineering territory. 

1. It removes the simplicity of NvChad and which, IMO, other configs (e.g., LunarVim & DoomNvim) have abandoned in favor of adding globs of glue code to make things "configurable", but needlessly adds abstractions.

 1. In addition to removing simplicity, it adds new surface area for bugs, bottlenecks, and maintenance down the road.

These changes snowball until you get stuff like [this](https://github.com/NTBBloodbath/doom-nvim/blob/main/lua/doom/core/config/init.lua) 600 line 'config' file which has thousands of glue code behind it.

I also wonder if there is a better way of checking configuration options than a conditional since they are considered a [code smell](https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3)

G-Rowell:
Hey @vladdoster,

What an insightful comment, for sure some good points there.


Ultimately there does have to be some plan for users to configure their own copy of NvChad,
Would you have any suggestions regarding this?

(I would argue it's not reasonable for every use to fork NvChad and fix their merge conflicts that result)


Certainly something to be cautious of,
If we streamlined these hooks & overrides into another table within `default_config` / `chadrc`, that could potentially reduce some complexity for the user?
(Though I know that doesn't completely takle what you're talking about)


This is certainly something that we've struggled with (to find a decent solution) and is honestly the last big hurdle for NvChad


G

osamuaoki:
FYI: If my PR to extension https://github.com/NvChad/extensions/pull/2 is applied, there will be no `git reset` worry so it will be OK to edit any place and you don't need to worry about erased.

Of course, we need to be prudent to pull upstream change but that can be accomodated by having another local copy of checkout to ensure clean rebase.

It will be interesting test case how my system behaves after this upcoming changes.

osamuaoki:
The existence of chadrc had 2 meanings.
* One was to make a concise point where user look for key reconfiguration.
    * One layer of variable assignment as a CPU cost
    * It offered a clean easily understandable table to be edited.  Good for human.
    * It eliminated chance of merge conflict for the related customization.
* One was to preserve user changes who tracks upstream repo directly.
    * This was a necessity coming from the current update script in extension.
    * I think we should allow user to edit other places and upgrade shouldn't reset them.  As I said above, my PR2 for extension enables this.

G-Rowell's code add one more layer of mechanism **override**.  I suppose he is thinking to offer similar override for other parts of configuration.  It's an interesting idea.

From user's POV, tracking down code to come up with proper lua table to override seems to be a lot of efforts.  I would rather edit things in the original assignment place.  I think this is unnecessary complication.  So I feel this is a bit of over engineering.

If this becomes the future NvChad, despite of the warning, I will be rather tempted to edit `default_config.lua` directly since I see things there more clearly.  (Of course, I will be making sure it isn't going to be erased by update script.)

As for configuring extra package etc., why not allow user to do so.  What's wrong editing `lua/plugins/init.lua`.   Changes to that part of code seems less likely to cause `git pull --rebase` conflict since modification will be line-wise as a insertion of block. iMy PR2  address `git reset` issue.

Of course, I am open to idea such as adding similar mechanism as what  "M.custom.mappings" offers for other parts of configuration in chadrc as long as it doesn't make user's configuration more complicated by offering commented out example there.

G-Rowell:
I think the fundamental problem with editing the general config is that there will almost certainly be merge conflicts.

Eg. 
- a user customises a plugin, and we remove it. That would indeed generate a merge conflict.
- a user changes a plugin option that we also change. That would be a merge conflict.

---

> One was to make a concise point where user look for key reconfiguration. 

In regards to this, I can re-work this PR slightly to have `hooks` & `overrides` take their values from a table in `chadrc`, doing so would centralise things and make this a cleaner solution I think.

I think this could be a great solution?

---

> , tracking down code to come up with proper lua table to override seems to be a lot of efforts.
 
In regards to this, proper documentation could be done to list all the overrides?
But chances are, if a user wants to modify a plugin, they will do two things:
1. read the plugins README
2. read through the NvChad config file `lua/plugins/configs/......`

Thus they will be able to also read, understand and modify using the overrides.

---


@ashincoder @osamuaoki 
Opinions?


G

ashincoder:
> I think the fundamental problem with editing the general config is that there will almost certainly be merge conflicts.
> 
> Eg.
> 
> * a user customises a plugin, and we remove it. That would indeed generate a merge conflict.
> * a user changes a plugin option that we also change. That would be a merge conflict.
> 
> > One was to make a concise point where user look for key reconfiguration.
> 
> In regards to this, I can re-work this PR slightly to have `hooks` & `overrides` take their values from a table in `chadrc`, doing so would centralise things and make this a cleaner solution I think.
> 
> I think this could be a great solution?
> 
> > , tracking down code to come up with proper lua table to override seems to be a lot of efforts.
> 
> In regards to this, proper documentation could be done to list all the overrides?
> But chances are, if a user wants to modify a plugin, they will do two things:
> 
> 1. read the plugins README
> 2. read through the NvChad config file `lua/plugins/configs/......`
> 
> Thus they will be able to also read, understand and modify using the overrides.
> 
> @ashincoder @osamuaoki
> Opinions?
> 
> G

I think these issue can be resolved. If some breaking changes are added we could create some kind of notice in the `doc` dir. And mention the users about it.

morten-olsen:
Can see the case for centralizing hooks in chadrc. The downside to this is that only one hook can be added to a given even.

For instance, in my setup https://github.com/morten-olsen/NvChad/tree/main/lua/custom i have multiple different plugins, which hooks into the plugin install and/or mapping, which allows me to have my customization in a way where I can easily add or remove an entire plugin, instead of having to track down its code multiple places in chadrc.

In terms of wanting to do the changes directly in `lua/plugins/init.lua` and dealing with the resulting merges, well that is still an option regardless if this hook system is there or not,. Anyone who want to manage NvChad as a fork and modify directly in the source can still do that.

G-Rowell:
> The downside to this is that only one hook can be added to a given even.

I don't think so?
We could have a table inside `chadrc`, with subvalues for each of the three hooks?

Eg:
in `chadrc`
```
M.custom.hooks = {
   mappings = {
      ....
   }
   m.autocmds = {
      ...
   }
}
```

G-Rowell:
Ultimately, with this PR we should decide on a semi-final format of custom config, which is then ideally **never changed**

Seems like some big points are:
1. avoiding complexity
2. preserve user custom config / changes
2. avoiding unnescasary `code hunting / investigating` (don't force the user to read through the code base too much)
3. if a breaking change is used, then label the commit as breaking, but avoid in the first place

siduck76:
> > The downside to this is that only one hook can be added to a given even.
> 
> I don't think so?
> We could have a table inside `chadrc`, with subvalues for each of the three hooks?
> 
> Eg:
> in `chadrc`
> 
> ```
> M.custom.hooks = {
>    mappings = {
>       ....
>    }
>    m.autocmds = {
>       ...
>    }
> }
> ```

looks good :smiley:

G-Rowell:
I'm thinking:

1. We can add functionality to custom config (IE. `chadrc`) in the future, we just **always have to backwards compatible**
    - adding functionality doesn't mean editing `chadrc`, just enhancing overrides & hooks in the `chadrc` file
2. We have to keep it simple & concise, as to avoid the bloat that @osamuaoki was mentioning

siduck76:
@G-Rowell Docs for explaining the overall config is very important too , at this point most issues I've seen were just caused by the user not knowing how nvchad's config works at all ;(

osamuaoki:
G-Rowell ,

I understand your concern for merge error.  I updated PR2 to ensure automatic roll back upon failed git pull --rebase.
This PR2 itself is much smaller change than PR1 for offset install in extension.  Once this updated git flow is used, any merge conflict is experienced, people who use this kind of tool should be able to safely resolve conflicts by making an independent check out.

The more I look at code, extra layer assignment adds more complication with little value for the trouble of user updating.  This is huge structural change which automatic merge resolver will not help.  (On the other hand, editing current code and resolving its conflict with upstream is much trivial.)  If you ever chase putting more capability to configure within chadrc, that should be very conservative for moving file etc.

Basically, we need to aim good engineering compromise.  That's how I see this.  So having I have no appetite to remove chadrc and edit every setting everywhere.  At the same time, I have no appetite to refactor and move files without significant gains.  If we focus more on smart package loading to make it work more efficiently, I am happy.  I was very impressed by the current code since it shortened my start up time below 70ms.

ashincoder
Notice to user is the last thing we want.  People don't want to read document or change set up for every iteration of update.  Best thing is things improve behind people realizing.  That's why I think improving updater to allow safer editing of any part of code is something important.

morten-olsen,
Editing plugins/init.lua is off-limit with the current extension logic.

If you use my PR2's no-reset update method, you can actually edit plugins/init.lua without issue.  Update will not erase it.  Even if in rare occasion, NvChad upstream decide to make non-mergeable changes, you will get proposed changes in git history only.

If PR2 https://github.com/NvChad/extensions/pull/2 needs to be made simpler by removing PR1 changes, I can create it easily.

Osamu

G-Rowell:
> Editing plugins/init.lua is off-limit with the current extension logic.

It's possible given this PR, with my `default_plugin_overrides` + Mortens hooks, I think

abalmos:
I wonder if this wouldn't be less complexity and still a decent term of tracking upstream:

1. Change updater from `git reset`(how rude!) to something more like `git stash && git pull --rebase && git stash pop`. If the rebase hits a conflict NvChad could give the user to quit and finish the rebase manually or just abort the update and issue a `git rebase --abort && git stash pop`. It _must_ be dealt with by the human at this point anyway. (NB: NvChad ought to encourage users to fork NvChad and commit their local changes. Maybe even have a wiki page for people to post their forks as inspiration to others--this would help sell the message that NvChad is intended to be a base config, or at least that is what I understand @siduck76 has intended it to be.)
2. Add `chadrc.lua` to `.gitignore` and have NvChad copy `default_config.lua` to `chadrc.lua` if it doesn't exist at runtime. Everyone gets the defaults out of the box and the file becomes truly a local config file. Newly added config options would have to be manually added to chadrc.lua to override the default value.
3. Switch the plugin structure from a list in `init.lua` which uses indirect calls to other files for mappings, configs, etc. to simply a "plugins" directory where each file is require'd in ls order during packing. Benefits: 
   -- Load order can still be preserved by filename `001Nvchad_extensions.lua`, `002nvim-lua_plenary.nvim.lua`, etc.
   -- _Everything_ having to do with a plugin stays in one simple file (mappings, config options, etc.) NvChad configured plugins can still require the chadrc setup for user convenience when it makes sense.
   -- Adding new plugins is easy: just drop a file into the plugins folder into your downstream fork.
   -- Sharing plugins is now easy: literally just steal the plugin file from another fork and drop it into yours
   -- Conflicts can only occur for plugins that are part of upstream _and_ user modified ... I don't think you can avoid this.
4. Remove `core` folder and promote it's contents up one level.
   -- All global options: `options.lua`
   -- All global mappings: `mappings.lua`
   -- All global autocmds: `autocommands.lua`
   -- All plugin specific things: `plugins/<plugin-name>.lua`

This removes all/most of the fancy redirection currently present and made worse in this PR (except the _simple_ table of values in chadrc) and makes NvChad much more approachable by new users.

I have had plans to fork NvChad into this over the last few weeks, but alas I have not found the time. I figured I would at least share it here for discussion sake.

vladdoster:
> I'm thinking:
> 
> 1. We can add functionality to custom config (IE. `chadrc`) in the future, we just **always have to backwards compatible**
>    
>    * adding functionality doesn't mean editing `chadrc`, just enhancing overrides & hooks in the `chadrc` file
> 2. We have to keep it simple & concise, as to avoid the bloat that @osamuaoki was mentioning

If #1 is considered acceptance criteria, then I would strongly suggest introducing [semantic versioning](https://semver.org/#semantic-versioning-200) to NvChad.

Backwards  compatible breaking changes would be major versions and make it easier to help user issues.

G-Rowell:
> If this is considered acceptance criteria, then I would strongly suggest introducing [semantic versioning](https://semver.org/#semantic-versioning-200) to NvChad.

This is great idea

G-Rowell:
@abalmos 

1. Agree, this is a good point, I think @osamuaoki has done some great work on this

--- 

2. Not sure if I agree here, I don't think you can `.gitignore` a file that is to be kept in the repo?
    - I experimented with this a while ago
    - Is it possible you are misunderstanding the current config? currently NvChad will combine `default_config` & `chadrc`, such that any values in `chadrc` overwrite `default_config`, thus users already get a good out of the box config with `default_config` 
3. Not sure if I agree with this either, with this PR (which does need to be improved a bit for sure), to modify an in-built plugin:
    - user can use a `chadrc` plugin option if it exists
    - users can use the [`override`s feature to modify a plugins config table](https://github.com/NvChad/NvChad/pull/465/files#diff-fd32d4fb2342b2664fde244a11b00dffe8bca604b5af780f3f752ecd5066c0b4R38-R53), it the override is implemented within the plugin config
        - (I need to add this functionality to several plugins, it [currently works for some LSP](https://github.com/NvChad/NvChad/pull/465/files#diff-9f2ea60dcf47caf192197bcff8823913b3b46441e32bb8b384d13b2a9385c67eR85), see this PR modified files)
    - LASTLY, if all else fails, users can totally take over the plugin config with [this chadrc table](https://github.com/NvChad/NvChad/pull/465/files#diff-d7459fe7bb90af4b87158080595101d717e87fd346695d1711bdb2795a467b79R19-R21), doing so [replaces the appropriate Packer call here](https://github.com/NvChad/NvChad/pull/465/files#diff-2b897160cb2d725c8147dbb975baa3f1ef193747258c9c3ecf3a0f5587e3fb49R64)

---

I also think there is a lot of benefit to having all of the mappings in one file, as opposed to having to hunt around in many files just to find what is mapped.

---

>  Load order can still be preserved by filename 001Nvchad_extensions.lua, 002nvim-lua_plenary.nvim.lua

I do really love this idea, not sure if we need it right now?

---

Users can add their own plugins with the proposed [install plugins hook](https://github.com/NvChad/NvChad/pull/465/files#diff-2b897160cb2d725c8147dbb975baa3f1ef193747258c9c3ecf3a0f5587e3fb49R328)

though this could be cleaned up & slightly improved though.

Morten has done this (and a lot of other work in this PR!), thanks to him! 😁

G-Rowell:
Wanna say that I greatly appreciate all of your comments & contributions here!

Definitely wanna get this merged, but also want it done right & proper.

All these comments have pointed out some great ways to improve this PR!

Thanks :)

abalmos:
@abalmos, thanks so much for all the great ideas, always love a good discussion here!

> Not sure if I agree here, I don't think you can .gitignore a file that is to be kept in the repo?
> I experimented with this a while ago
> Is it possible you are misunderstanding the current config? currently NvChad will combine default_config & chadrc, such that any values in chadrc overwrite default_config, thus users already get a good out of the box config with default_config

That is my point though. `chadrc` should not be in git at all. That is, by definition, my file to modify not NvChad's. I don't know of any projects which track their configs in git other then defaults (in this case `default_config`) (...I'm sure you can find some if you try...). 

What I am suggesting is to remove `chadrc` from git, add it to the `.gitignore`, and then write code that copies `default_config` to `chadrc` at runtime if it does not exist. NvChad the project manages `default_config`, I as the user manages `chadrc`. You get the same out-of-box as now but none of the issues surrounding future commits.

> 2. Not sure if I agree with this either, with this PR (which does need to be improved a bit for sure), to modify an in-built plugin:
>    
>    * user can use a `chadrc` plugin option if it exists, IE. both are used
>    * users can use the [`override`s feature to modify a plugins config table](https://github.com/NvChad/NvChad/pull/465/files#diff-fd32d4fb2342b2664fde244a11b00dffe8bca604b5af780f3f752ecd5066c0b4R38-R53), it the override is implemented within the plugin config
>      * (I need to add this functionality to several plugins, it [currently works for some LSP](https://github.com/NvChad/NvChad/pull/465/files#diff-9f2ea60dcf47caf192197bcff8823913b3b46441e32bb8b384d13b2a9385c67eR85), see this PR modified files)
>    * LASTLY, if all else fails, users can totally take over the plugin config with [this chadrc table](https://github.com/NvChad/NvChad/pull/465/files#diff-d7459fe7bb90af4b87158080595101d717e87fd346695d1711bdb2795a467b79R19-R21), doing so [replaces the appropriate Packer call here](https://github.com/NvChad/NvChad/pull/465/files#diff-2b897160cb2d725c8147dbb975baa3f1ef193747258c9c3ecf3a0f5587e3fb49R64)

Right, but I believe my suggestion (which would be an alternative to all the types of hooks) gives the same benefit. Actually, this method has one big downside IMHO: when I take over a plugin fully, I have two copies of that plugins config. The one that comes from upstream but is now ignored at runtime and my version. 6 months from now I'll forgot what ones I overrode.

With my suggestion NvChad needs very little code (a for loop over files), I can edit plugins at will, and if I need to make large agricultural changes I can just drop a new file in. If you rename a file not `.lua` it wont be loaded at runtime and when I do a `git pull --rebase` my rename will survive any upstream changes.

> I also think there is a lot of benefit to having all of the mappings in one file, as opposed to having to hunt around in many files just to find what is mapped.

We'll just have to disagree then :). Right now I got to the plugin file to find its in the mapping file to find its in chadrc. I'm not the author so I don't off hand where everything is.
 
>>Load order can still be preserved by filename 001Nvchad_extensions.lua, 002nvim-lua_plenary.nvim.lua
>I do really love this idea, not sure if we need it right now?

Only if this went the direction I suggest.


Sorry for the rushed response, but I have run out of time for now. If you find any merit I am happy to keep explaining.

G-Rowell:
@abalmos 
Really appreciate your detailed response!!!


It's 1am here so I'll have to come back tmr,
Just wanted to say thanks, 😃

siduck76:
Ok now Im actually gonna look at this PR! Btw this looks bad 

image

should they do this everytime they update nvchad?

siduck76 (commented):
also dont include UI in options table :l

lua/core/default_config.lua
```diff
@@ -0,0 +1,221 @@
...
+   undofile = true, -- keep a permanent undo (across restarts)
+   -- NvChad options
+   nvChad = {
+      copy_cut = true, -- copy cut text ( x key ), visual and normal mode
+      copy_del = true, -- copy deleted text ( dd key ), visual and normal mode
+      insert_nav = true, -- navigation in insertmode
+      window_nav = true,
+      theme_toggler = false,
+      -- used for updater
+      update_url = "https://github.com/NvChad/NvChad",
```
siduck76, line 37: as mentioned in the description of this PR that the updater.lua will no longer work , so why is this update url needed?

lua/core/default_config.lua
```diff
@@ -0,0 +1,221 @@
...
+      update_branch = "main",
+   },
+}
+
+-- ui configs
+M.options.ui = {
```
siduck76, line 43: UI must be a different thing imo ( M.ui ) , I'll add many UI stuffs in the future as well so separating it from options table would be good thing

G-Rowell:
> Ok now Im actually gonna look at this PR! Btw this looks bad
> 
> image
> 
> should they do this everytime they update nvchad? it gets cumbersome to keep having backups everytime we update nvchad :/

No, this is only when the user first updates to this PR.

This is because this PR will break things, so we should suggest that users backup their `chad` & custom configs, then `git fetch --all` `git reset origin/HEAD --hard`

This will completely remove all modifications, then they can copy paste them back into the new `chadrc` & custom section

lua/core/default_config.lua
G-Rowell: Not exactly,

`updater.lua` will need to be slightly changed, as we are moving `chadrc` to lua/custom/`.

this is why.

siduck76:
> > Ok now Im actually gonna look at this PR! Btw this looks bad
> > image
> > should they do this everytime they update nvchad? it gets cumbersome to keep having backups everytime we update nvchad :/
> 
> No, this is only when the user first updates to this PR.
> 
> This is because this PR will break their `chadrc`, so we should suggest that users backup their `chadrc` & custom configs, then `git fetch --all` `git reset origin/HEAD --hard`
> 
> This will completely remove all modifications, then they can copy paste them back into the new `chadrc` & custom section

oh cool then!

danjenson:
gotta say...I'm watching this diff with bated breath...stoked for it.

lua/core/default_config.lua
G-Rowell: Done,
you have a good point if there are a lot of UI feats added in the future

siduck76 (approved):

lua/plugins/configs/statusline.lua
```diff
@@ -42,7 +42,7 @@ local icon_styles = {
    },
 }
 
-local user_statusline_style = require("core.utils").load_config().ui.plugin.statusline.style
+local user_statusline_style = require("core.utils").load_config().plugins.options.statusline.style
```
siduck76, line 45: change it back to ui

vladdoster:
This is a breaking change and should be considered a major version, such as `v2.0.0`.

siduck76:
> This is a breaking change and should be considered a major version, such as `v2.0.0`.

agreed! please dont update nvchad yet :c

PR #476, opened 2021-09-17
ignore node_module folder in telescope

ayushjaipuriyar:
Just a minor tweak to telescope config

1. Added node_modules to file_ignore_patterns

It's just annoying to see the node_modules folder in telescope, and it doesn't always match correctly when there are too many files.

### DIFF
--- a/lua/plugins/configs/telescope.lua
+++ b/lua/plugins/configs/telescope.lua
@@ -35,7 +35,7 @@ telescope.setup {
          preview_cutoff = 120,
       },
       file_sorter = require("telescope.sorters").get_fuzzy_file,
-      file_ignore_patterns = {},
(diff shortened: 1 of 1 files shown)


G-Rowell:
Hmmmm, 

while I do think this is useful, 
it's definitely a language specific change.

This should probably use part of the upcoming #465 custom config overhaul.

ayushjaipuriyar:
> Hmmmm,
> 
> while I do think this is useful,
> it's definitely a language specific change.
> 
> This should probably use part of the upcoming #465 custom config overhaul.

Kinda language-specific, but I don't think it belongs in the custom configs because there's a one-in-a-million chance that someone will mess with the modules.

siduck76:
@G-Rowell nvim tree does it too 

image


I suppose this is someting important!

PR #493, opened 2021-09-26
fix: Replace Website Link in README.md

marvelman3284:
With the new docs I changed the links in the Readme. The link in the about section of the repo also needs to be updated, but its not something I can do.


G-Rowell:
Ty!

I like this

PR #496, opened 2021-09-26
fix: ignore unwanted files in custom dir

chebro:
- since it's pointless to track user specific files, added the custom dir to the ignore list except for the example files.


G-Rowell:
This is a great idea!

Thanks for the PR 😁

PR #497, opened 2021-09-26
Moved the setting into setup for Nvimtree

macarthy:
Options are currently being migrated into the setup function, you need to run require'nvim-tree'.setup() in your personal configurations. Setup should be run in a lua file or in a lua heredoc (:help lua-heredoc) if using in a vim file. Note that options under the g: command should be set BEFORE running the setup function.

I've moved the ones that have been migrated

PR #498, opened 2021-09-26
chore(nvim-tree): Move some settings to setup

ashincoder:
In this PR some settings like width, side have been moved to the setup. And the keymappings are removed cause they are by default. So why bother adding.


macarthy:
I have a branch submitted on the same issue.  But you are right about the settings if the are exactly the same 

Are you missing allow_resize = true,  or is that true by default ?

macarthy:
Also I was getting an error with Just doing more testing and seem to have a issue with
the below key binding, is it working on your commit ? 
 { key = "g?", cb = tree_cb "toggle_help" }

siduck76 (commented):
@ashincoder  https://github.com/NvChad/NvChad/pull/497

ashincoder:
> I have a branch submitted on the same issue. But you are right about the settings if the are exactly the same
> 
> Are you missing allow_resize = true, or is that true by default ?

Oh yeah just saw that. I just saw the change in nvim-tree and just made it up.

PR #510, opened 2021-09-30
fix: Some fixes for sept cleanup !

ashincoder:
1. Show line diagnostics have been changed to `ge`. Since there are keybindings starting with g. Also `ge` is fast very useful for this reason.

2.Making an autocmd for nvimtree can slow down the process and lag since the refresh autcmd in not asynchorous.

3. Fixes the issue of not opening a dir/folder. Fixed by enabling netrw only not other netrw settings. Netrw works without other plugins settings.Since it solves the issue moreover it doesn't affect the startuptime by any millisecond. Atleast for me. (Ran 5 tests). So win win.

:smile: :rocket: :100:


ashincoder:
Please make a dev branch it is very important. Compulsory.

siduck76 (approved):

siduck76:
@ashincoder you have commented netrw stuff because nvimtree already disables it right?

ashincoder:
we should enable netrw. cause nvimtree will not open in startup since we are lazyloading it. So our only options are :
1. Enable netrw and lazyload nvimtree. (I prefer this one since netrw is simple and lightweight compared to nvimtree which has many icons and git integration.)
2. Do not lazyload nvimtree, so it can open a dir.

PR #512, opened 2021-09-30
Create keyssheet_nvchad.md

Giacomo002:
Using chadwm and NvChad just created a keys sheet for both.


xluffy:
+1

siduck76:
@Giacomo002 we're considering adding a cheatsheet plugin or creating one after v2.0 release! for the time being you could add this in the docs 

here https://github.com/NvChad/nvchad.github.io/blob/src/docs/config.md?plain=1#L652

PR #514, opened 2021-10-01
Fixes/september cleanup

ashincoder:
Show line diagnostics have been changed to ge. Since there are keybindings starting with g. Also ge is fast very useful for this reason.
2.Making an autocmd for nvimtree can slow down the process and lag since the refresh autcmd in not asynchorous.

Fixes the issue of not opening a dir/folder. Fixed by enabling netrw only not other netrw settings. Netrw works without other plugins settings.Since it solves the issue moreover it doesn't affect the startuptime by any millisecond. Atleast for me. (Ran 5 tests). So win win.
smile rocket 100


siduck76:
enabling netrw isnt a good idea imo

image

ashincoder:
> enabling netrw isnt a good idea imo
> 
> image

choose between option 1 2, & 3. Straight answer only !.

ashincoder:
we should enable netrw. cause nvimtree will not open in startup since we are lazyloading it. So our only options are :

1. Enable netrw and lazyload nvimtree. (I prefer this one since netrw is simple and lightweight compared to nvimtree which has many icons and git integration.)
2. Do not lazyload nvimtree, so it can open a dir.
3. Let the issue of not opening a dir continue.

ashincoder:
> enabling netrw isnt a good idea imo
> 
> image

I don't know why u say that it does little change to the startuptime.
(meaning of little: not at all [used before a verb])

G-Rowell:
@ashincoder 

I've been tinkering with a way to keep `netrw` disabled, while allowing NvimTree to open on `vim <DIR>`.

Made some good progress, and personally I'd prefer to only have 1 tree plugin enabled.

I don't think NvimTree is too heavy as well?


What's your opinion?

G-Rowell:
> 
> 
> we should enable netrw. cause nvimtree will not open in startup since we are lazyloading it. So our only options are :
> 
>     1. Enable netrw and lazyload nvimtree. (I prefer this one since netrw is simple and lightweight compared to nvimtree which has many icons and git integration.)
> 
>     2. Do not lazyload nvimtree, so it can open a dir.
> 
>     3. Let the issue of not opening a dir continue.



I am quite happy to go with option 2,

I am working on making option 3 possible
(lazy load NvimTree but still have it be able to open dirs)

ashincoder:
> > we should enable netrw. cause nvimtree will not open in startup since we are lazyloading it. So our only options are :
> > ```
> > 1. Enable netrw and lazyload nvimtree. (I prefer this one since netrw is simple and lightweight compared to nvimtree which has many icons and git integration.)
> > 
> > 2. Do not lazyload nvimtree, so it can open a dir.
> > 
> > 3. Let the issue of not opening a dir continue.
> > ```
> 
> I am quite happy to go with option 2,
> 
> I am working on making option 3 possible (lazy load NvimTree but still have it be able to open dirs)

nvimtree is quite heavy based on the git integration. git is not a small program. Tinkering with it can cause some delay. Whilst netrw is written in vimscript and c. So it is binded together.

ashincoder:
> @ashincoder
> 
> I've been tinkering with a way to keep `netrw` disabled, while allowing NvimTree to open on `vim <DIR>`.
> 
> Made some good progress, and personally I'd prefer to only have 1 tree plugin enabled.
> 
> I don't think NvimTree is too heavy as well?
> 
> What's your opinion?

Also netrw will only be opened when in a directory. I would love to have option lazyloading for nvimtree with also opening in startup.

G-Rowell:
> 
> 
> > @ashincoder
> > I've been tinkering with a way to keep `netrw` disabled, while allowing NvimTree to open on `vim <DIR>`.
> > Made some good progress, and personally I'd prefer to only have 1 tree plugin enabled.
> > I don't think NvimTree is too heavy as well?
> > What's your opinion?
> 
> Also netrw will only be opened when in a directory. I would love to have option lazyloading for nvimtree with also opening in startup.


Let me tinker a little more then!

I personally would be more then happy to just not lazy-load it, what are your thoughts on this?

G-Rowell:
Though I would like to add that `NvimTree` has done & is still improving it's speed.

So it may be faster/less impactful these days?

Still will work on auto open + lazy loading

siduck76 (approved):

PR #541, opened 2021-10-08
fix: new function to check for cmp completitions

max397574:
https://github.com/hrsh7th/nvim-cmp/issues/231


siduck76:
did you test it? pressing TAB gives me some error

ashincoder:
@siduck76 do :PackerSync

siduck76:
> Does anyone know if I can make that window transparent?

just make those  relevant highlight's guibg=NONE

PR #544, opened 2021-10-09
fix: use of left/right arrow keys in insert mode

Ay-355:
This allows the left and right arrows to move between lines in insert mode. Earlier it only worked in normal mode.


ashincoder:
@Ay-355 could u do a demonstration in this change. I can't work out what this does.

PR #545, opened 2021-10-09
update nvim-cmp plugin

thuan1412:
Follow [this](https://github.com/hrsh7th/nvim-cmp/issues/231) breaking change of `nvim-cmp`

We need to run `PackerUpdate` after updating this config.


ashincoder:
@thuan1412 implemented.

PR #596, opened 2021-11-08
added highlight override support

zbirenbaum:
I use classic vim-tabs, but highlights do not function no matter what I do. Something like vim.cmd[[hi TabLineFill "#000000"]] works flawlessly when I :tabe if I run it from wildmenu with lua, but does not function at all, even if run with an au group like ColorScheme or BufEnter when placed elsewhere in my config. I have found out it is being overwritten by the colorscheme code at some point, as the only location in which it works is if I put it at the bottom of lua/colors/highlights.

To my knowledge, this is not available for override specifically, and adding a default config override the colors.init() call in the config portion of nvim-base16 would require copying over the entirety of the colors folder into custom for like 4 lines of code at the bottom of highlights.

An example of the file I used to test the functionality of my code is as follows:
first, I set hl_override in chadrc to "custom.plugins.hl_override:

inside custom.plugins.hl_override.lua:
```
local cmd=vim.cmd

local function bg(group, color)
   cmd("hi " .. group .. " guibg=" .. color)
end

local function fg(group, color)
   cmd("hi " .. group .. " guifg=" .. color)
end

bg("TabLineFill", "#000000")
```
Result: everything works fine. Setting hl_override to false does not call the file in question and everything proceeds normally. Thus, I do not feel a pcall/present check is necessary, since the user must supply a specific path of their own, and we want them to be notified it cannot be found, thus the require statement in highlights.lua.

### DIFF
--- a/lua/colors/highlights.lua
+++ b/lua/colors/highlights.lua
@@ -1,3 +1,4 @@
+local override = require("core.utils").load_config().options.hl_override
 local cmd = vim.cmd
 
 local colors = require("colors").get()
@@ -155,3 +156,7 @@ fg("TelescopeBorder", one_bg)
 fg_bg("TelescopePreviewTitle", green, one_bg)
 fg_bg("TelescopePromptTitle", blue, one_bg)
 fg_bg("TelescopeResultsTitle", red, one_bg)
+
+if override then
+   require(override)
+end
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -38,6 +38,7 @@ M.options = {
       update_url = "https://github.com/NvChad/NvChad",
       update_branch = "main",
    },
+  hl_override = false, --set this equal to "custom.some.file" (which contains highlights) in custom.chadrc to modify hl groups set by the colorscheme
 }
 
 -- ui configs


siduck:
good idea but this wont work 

image

siduck:
also move the variable in the ui table

siduck:
nvm it , i'll make this change in a single commit. Thanks for this tho!

zbirenbaum:
> good idea but this wont work 
> 
> 
> 
> image
> 
> 

How did you produce this? I never got it to error unless I typed in a non existent file. I can move it over too np

zbirenbaum:
> nvm it , i'll make this change in a single commit. Thanks for this tho!

Nooooo my chance at a contributor tag 😭. Jk, thanks for the quick response,this bug really drove me wild.

siduck:
@zbirenbaum The change is done ! https://github.com/NvChad/NvChad/commit/fd668e559ed6ccfdd565e3f43a7ae1c6efc16582

siduck:
> > nvm it , i'll make this change in a single commit. Thanks for this tho!
> 
> Nooooo my chance at a contributor tag sob. Jk, thanks for the quick response,this bug really drove me wild.

no worries you can always contribute to the v2.0 release ( check the pinned issue) . I lost all of my contributions when I renamed my username from siduck76 to siduck 

before : 

image

zbirenbaum:
> > > nvm it , i'll make this change in a single commit. Thanks for this tho!
> > 
> > 
> > Nooooo my chance at a contributor tag sob. Jk, thanks for the quick response,this bug really drove me wild.
> 
> no worries you can always contribute to the v2.0 release ( check the pinned issue) . I lost all of my contributions when I renamed my username from siduck76 to siduck 😆
> 
> before :
> 
> image
> 
> now :
> 
> image

Once classes lighten up a bit I'll be sure to get on that! I've seen it open for a while and since NvChad has long become the basis I do my customizations off of I have a vested interest in seeing it improve even more lmao.

Ay-355:
@siduck your losing commits problem seems to be caused by your email change. As far as I know, changing your username doesn't remove contributions. [Says here that you can add the email back to not lose any contributions](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-user-account-settings/best-practices-for-leaving-your-company#:~:text=You%20can%20then%20re-add%20it%20without%20verifying%20to%20keep%20any%20associated%20commits%20linked%20to%20your%20account.)

siduck:
> @siduck your losing commits problem seems to be caused by your email change. As far as I know, changing your username doesn't remove contributions. [Says here that you can add the email back to not lose any contributions](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-user-account-settings/best-practices-for-leaving-your-company#:~:text=You%20can%20then%20re-add%20it%20without%20verifying%20to%20keep%20any%20associated%20commits%20linked%20to%20your%20account.)

yep , forgot to mention that xD

siduck:
@Ay-355 i added the email back but my contribution count shows just to me , doesnt work in incognito.

image
image

siduck:
image


D:

Ay-355:
Ah that's weird, I had the same problem and re-adding the email fixed it. Maybe they'll turn up later

Ay-355:
And will you look at that, they ended up coming back!

PR #662, opened 2021-12-18
Make lspconfig mappings configurable.

horta:


siduck:
I doubt if the map function will work here 

image

check

siduck:
@horta whenever you add a table to the mappings table , it gets added to the cheatsheet :D

image

PR #673, opened 2021-12-23
fix: change nvim-tree open directory icon

DanielCardonaRojas:
Just changes the icon for an open folder, might be opinionated but I think the expected behaviour would be that: 

image


siduck:
i had given a thought to this before but the open icon didnt match with the folder icon in size so i didnt use it

PR #700, opened 2022-01-03
support update default lsp key mappings

zhaopengme:


siduck:
I don't think the hooks function is necessary here

siduck:
hooks have been removed! 

https://github.com/NvChad/NvChad/commit/4e54ee0081babd3138180bb61ea093fcac03fd46

PR #702, opened 2022-01-03
fix errors thrown by implicit conversions in latest neovim build

zbirenbaum:
neovim/neovim#16745 which was merged into master a bit ago causes errors on opening nvchad due to no longer accepting nil in numerous api functions. The statusline config made use of this implicit conversion and I changed it to convert nil to 0 prior to function call.  A side effect of the implementation of the neovim commit is that '%' no longer can be used to refer to current buffer. This may be changed slightly later as I have seen a few commits over the last few hours updating related functions elsewhere in neovim, but I wouldn't count on it and these edits are necessary to keep nvchad running on up to date builds, and are the 'correct' implementation regardless of what other inputs they decide to reallow.

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -22,6 +22,7 @@ M.close_buffer = function(force)
    local function switch_buffer(windows, buf)
       local cur_win = vim.fn.winnr()
       for _, winid in ipairs(windows) do
+         winid = tonumber(winid) or 0
          vim.cmd(string.format("%d wincmd w", vim.fn.win_id2win(winid)))
          vim.cmd(string.format("buffer %d", buf))
       end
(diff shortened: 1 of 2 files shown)


peppapig450:
Isn't this on the nightly build? And nightly doesnt work with telescope anyways

zbirenbaum:
> Isn't this on the nightly build? And lot of other plugins like telescope don't work on nightly either

Yes it is, but many NvChad users use nightly. I personally don't use telescope for much so I can't comment as to its functionality, but what I can say is that regardless of whether telescope actually functions or not (I do have it enabled in my config) neovim doesn't throw 100 error messages preventing me from editing a file on startup due to whatever incompatibilities it may have. Unfortunately, without the changes in this PR, NvChad does do just that.

siduck:
the change is included in the master only right? or 0.6 too?

peppapig450:
This works on v0.6.1 too I just tested it on both nightly and stable

PR #717, opened 2022-01-08
when finding all files via Telescope follow symlinks

starkers:
Hey there,

first day chad user here, this is just a small improvement, hopefully I'll be able to figure out a more elegant way to allow users to tweak their mappings soon.

Really enjoying the layout, its got its problems but its brilliant IMO.. thanks

Dunno how you feel on the subject but I'd suggest adding this to the main `find_files` also

Ref: https://github.com/nvim-telescope/telescope.nvim/blob/f06dd06bb1143caa779e492ca37e5f985f0c6157/doc/telescope.txt#L768

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -169,7 +169,7 @@ M.telescope = function()
 
    map("n", m.buffers, ":Telescope buffers <CR>")
    map("n", m.find_files, ":Telescope find_files <CR>")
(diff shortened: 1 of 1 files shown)


siduck:
tweaking mappings is  easy! check this 
image

starkers:
ye, its really hit and miss tbh and I've not figured out why yet

image


wrong results still

image

I've sometimes taken to hacking them in via pure `.vim` but its probably me

siduck:
image

chadrc : 
image

PR #724, opened 2022-01-11
Statusline Rewrite for Clarity and Customizability

zbirenbaum:
A while ago I wanted to override the default NvChad statusline with a couple of changes, and quickly found that adding any new component, or removing any component would require changing the number in the insert function of every subsequent component in the corresponding area of the statusline, namely being the left (1), middle(2) -but only one component so non-issue really, and right(3). Now there are quite a few components in the NvChad statusline and this quickly became extremely cumbersome. While my statusline now differs greatly from NvChad, the flexibility of my method has been invaluable in customizing it for clarity and time saving purposes so I felt updating it here might be a benefit to many individuals.

The NvChad statusline is appealing to many users and has some functionality that is likely non-trivial for inexperienced or even possibly some intermediate users to implement on their own, and thus turning it into something which is a bit more 'hacker friendly' might be a great benefit to some users. I know that for many of my overrides, especially ones written before I became more comfortable with lua, started with the NvChad configuration and go from there.

My changes create a function which returns a table of named components. Some components which logically belong in groups (like the 3 git diff ones or the 4 diagnostic ones) have been grouped into a nested table such that components_list.diagnostics.errors refers to the error component for instance, while lsp progress and lsp as more isolated components are simply referred to as components_list.lsp_progress and components_list.lsp respectively for their insertion

Furthermore, 1,2, and 3 have been assigned the values of left, middle, and right so that it is clear in the insertion where they are going. All insertions take place near the bottom of the code, so refer there to see the updated methodology. A massive benefit of this approach is that changing the order of components, or inserting a new one is as simple as changing the sequential order of the insertions or placing a new one between the components you wish to place it. Removal is equally easy by just taking one out. In the current functionality, removing the 'directory' component would require decrementing every subsequent left component from components.active[1][4] -> components.active[1][10] to be components.active[1][3] -> components.active[1][9].

in my implementation it is as simple as making the following comment:
```
table.insert(left, components_list.main_icon)
table.insert(left, components_list.file)
--table.insert(left, components_list.dir)
table.insert(left, components_list.git.added)
table.insert(left, components_list.git.modified)
table.insert(left, components_list.git.removed)
table.insert(left, components_list.diagnostics.errors)
table.insert(left, components_list.diagnostics.warnings)
table.insert(left, components_list.diagnostics.hints)
table.insert(left, components_list.diagnostics.info)
```

All functionality and display seemed to be the exact same as when I used a vanilla NvChad config in my tests but I would greatly appreciate someone else testing this and giving confirmation as it is possible I missed something.

Personally I like having the components in a separate file and getting them through a require, but since the original implementation was in one file I kept mine in line with that. I am happy to change it to be in a folder. I believe if the statusline 'building' portion is called init.lua and the components place in a file called components.lua putting them in a directory should not break anything elsewhere in the code, but I have not tested that yet so I am not certain.

### DIFF
--- a/lua/plugins/configs/statusline.lua
+++ b/lua/plugins/configs/statusline.lua
@@ -3,44 +3,44 @@ local lsp = require "feline.providers.lsp"
 local lsp_severity = vim.diagnostic.severity
 
 local icon_styles = {
-   default = {
-      left = "",
-      right = " ",
-      main_icon = "  ",
-      vi_mode_icon = " ",
-      position_icon = " ",
-   },
-   arrow = {
-      left = "",
-      right = "",
-      main_icon = "  ",
-      vi_mode_icon = " ",
-      position_icon = " ",
-   },
+	default = {
+		left = "",
+		right = " ",
+		main_icon = "  ",
+		vi_mode_icon = " ",
+		position_icon = " ",
+	},
+	arrow = {
+		left = "",
+		right = "",
+		main_icon = "  ",
+		vi_mode_icon = " ",
+		position_icon = " ",
+	},
 
-   block = {
-      left = " ",
-      right = " ",
-      main_icon = "   ",
-      vi_mode_icon = "  ",
-      position_icon = "  ",
-   },
+	block = {
+		left = " ",
+		right = " ",
+		main_icon = "   ",
+		vi_mode_icon = "  ",
+		position_icon = "  ",
+	},
 
-   round = {
-      left = "",
-      right = "",
-      main_icon = "  ",
-      vi_mode_icon = " ",
-      position_icon = " ",
-   },
+	round = {
+		left = "",
+		right = "",
+		main_icon = "  ",
+		vi_mode_icon = " ",
+		position_icon = " ",
+	},
 
-   slant = {
-      left = " ",
-      right = " ",
-      main_icon = "  ",
-      vi_mode_icon = " ",
-      position_icon = " ",
-   },
+	slant = {
+		left = " ",
+		right = " ",
+		main_icon = "  ",
+		vi_mode_icon = " ",
+		position_icon = " ",
+	},
 }
 
 local config = require("core.utils").load_config().plugins.options.statusline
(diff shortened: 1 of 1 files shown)


siduck:
ngl this looks more complex than the previous statusline conf!

siduck:
i'll clean the statusline config later

zbirenbaum:
@siduck all this did was put each component into a table and write a function which returns the table. The only thing which really changed was the entry process which has been simplified to as series of table.inserts(). I fail to see how it is any more complex, and I can't think of a better way to do this without retaining the issue of being unable to change things. Could you re-open this and take another look? it might be helpful if you look at it in terms of two separate files like I have for mine:

Here is the actual status line building
https://github.com/zbirenbaum/NvCustom/blob/remove_hooks/plugins/overrides/statusline_builder/builder.lua
Here are the components
https://github.com/zbirenbaum/NvCustom/blob/remove_hooks/plugins/overrides/statusline_builder/components.lua

I have had my statusline overridden for a really long time but I thought it would be really helpful for other people to have this setup so I spent a decent amount of time refactoring it and I'd really appreciate it if you could take another look.

siduck:
@zbirenbaum it was great but too much nesting :(

zbirenbaum:
If I remove the nesting and re-PR would that be acceptable?

siduck:
> If I remove the nesting and re-PR would that be acceptable?

just replace this 

https://paste.sh/tUBS3eId#bOMivCqj066EfSEIjnuoqBUm

siduck:
nvm i'll do it myself!

siduck (approved):

zbirenbaum:
Lol we did it at the exact same time, I just finished removing the nesting on my end. Thanks for taking another look at it, hope my work helped a bit!

siduck:
> Lol we did it at the exact same time, I just finished removing the nesting on my end. Thanks for taking another look at it, hope my work helped a bit!

hehe almost! but you used M.stuff  and function which didnt look good imo, anyways thanks for the PR. really helped in organizing the statusline config :)

zbirenbaum:
Ahh yeah you used local vars. Your implementation is much clearer for a single file config now that I look at it, I didn't event think about doing that. Good implementation and happy to help :)

Issue #620, opened 2021-11-30
Add/Remove/Overwrite individual keys in the NvChad Plugins Setup tables?


erwin:
**Is your feature request related to a problem? Please describe.**

What do you guys think about allowing each key of the plugin setup table to be overridden/added/removed, instead of entirely replacing the setup table the way you currently document?

```
M.plugins = {
   default_plugin_config_replace = {
      lspconfig = "custom.lspconfig",
   },
}
```

For example, I wanted to change add a few `sources` to the `cmp.lua` config. Instead of a way to append or remove from the cmp.setup.sources map, I had to make a copy of `cmp.lua` into `lua/custom/plugins/cmp.lua` and duplicate the whole file and make the changes.

```
   sources = {
      { name = "nvim_lsp" },
      { name = "luasnip" },
      { name = "buffer" },
      { name = "nvim_lua" },
      { name = "path" },
      { name = "neorg" },
   },

```

Similarly for bufferline.lua I wanted to add ordinal tab numbers, and had duplicate the all 143 lines of the bufferline file just to add a `numbers` function.

```
      numbers = function(opts)
        return string.format(
          '%s.%s', opts.lower(opts.id), opts.lower(opts.ordinal))
      end,
```

**Describe the solution you'd like**

It's not a big deal. NvChad works fine as it is.
In some cases, like `lsp-config` I think it's really better to just overwrite the whole thing.

Just wondering there's a way PackerCompile might be used so that an expensive operation like merging each key of a user provided table with each of the `NvChad` provided setup tables could be done without sacrificing any of the NvChad config's speed.

**Describe alternatives you've considered**
Just a thought. It's surely already possible to get the same result. Just would like to do something more granular, as for NvChad users that should be easier to maintain over time.

siduck:
PR's are welcome! it'll take a while to add this stuff sincemy college is going on ;(

zbirenbaum:
I made something like this for feline the other day when I was completely rewriting it and wanted the ability to swap around components fast. I would PR it but I completely changed every component and don't have time to make a version for the base config rn. I may work on it and others in a week or two after finals, but college be rough this time of year for everyone I think. gl @siduck ace those finals bro

siduck:
there are tons of plugins and adding tables for each and every plugin would populate chadrc and make it more lengthy un-necessarily, the user can just copy default plugin config and edit it rather than us adding more code in the config which is not needed , dont be that lazy!! :angry:

[...]

PR #735, opened 2022-01-21
Add/Remove/Overwrite individual keys in NvChad setups

zbirenbaum:
# Implements support for #620: Add/Remove/Overwrite individual keys in the NvChad Plugins Setup tables

### The following plugins (with their corresponding chadrc override key in parenthesis  for convenience) support table overrides:

- bufferline.nvim (bufferline)
- comment.nvim (nvim_comment)
- gitsigns (gitsigns)
- indent-blankline.nvim
- LuaSnip (luasnip)
- lsp_signature (signature)
- nvim-autopairs (nvim_autopairs)
- nvim-cmp (nvim_cmp)
- nvim-colorizer (nvim_colorizer)
- nvim-lspconfig (nvim_lspconfig)
- nvim-tree (nvim_tree)
- nvim-treesitter (nvim_treesitter)
- nvim-web-devicons (nvim_web_devicons)
- Telescope (telescope)

### The following plugins do not support table overrides but all defaults are configurable via chadrc's options section

- better-escape
- dashboard-nvim

### The following plugins do not make use of unified setup tables like the others, and thus do not support table overrides and require their own configuration files to change their behavior:

-  lspconfig
-  lsp_handler

### Examples of usage for various plugins:
_**Note: See the nvim_cmp section of the below example for how to override lspkind_icons as part of cmp**_
```
default_plugin_config_replace = {
   telescope = { defaults = { prompt_prefix = " ?  "}},
   nvim_treesitter = { highlight = { enable = true}},
   nvim_autopairs = {},
   nvim_web_devicons = {
      override ={
         lua = {
            icon = "L",
            name = "lua",
         },
      }
   },
   gitsigns = {
      signs = {
         add = { hl = "DiffAdd", text = ">", numhl = "GitSignsAddNr" },
      }
   },
   nvim_colorizer = {  --foreground should be black instead of background: #000000
      user_default_options = {
           mode = "foreground"
      },
   },
   bufferline = { options= {buffer_close_icon = "+"}},

   signature="custom.plugins.overrides.cmp_configs.lspsignature_cmp",
   feline="custom.plugins.overrides.statusline_builder.builder",
   
   nvim_cmp = { --To override icons the rest of the format function must be copied here as well
      formatting = {
         format = function(entry, vim_item)
            local icons = require "plugins.configs.lspkind_icons"
            
            _**--The following line edits lspkind_icons and is not from the original config**_
            icons = vim.tbl_deep_extend("force", icons, { Text="[]", TypeParameter="[]", Unit="[]"})
            
            vim_item.kind = string.format("%s %s", icons[vim_item.kind], vim_item.kind)

            vim_item.menu = ({
               nvim_lsp = "[LSP]",
               nvim_lua = "[Lua]",
               buffer = "[BUF]",
            })[entry.source.name]

            return vim_item
         end,
      },
   }
},
```
### Implementation notes and reasonings

The changes to plugin configurations to return a table and then be called via setup() (or setup(true) if overridden) from init.lua rather than just requiring the module are necessary since any implementation which does not return a table (like the ones mentioned prior in #620) involve adding some level of runtime overhead even for those not taking advantage of the feature (which out of principle I would refuse to implement, as were such a change merged I would make a fork removing it). You can easily tell that the initial implementations had runtime implications as any change made in the way shown prior in #620 would take effect on restart, while adding a new configuration file requires packer to recompile to update. 

The implementation in this PR maintains runtime efficiency and consistency, as it is only on recompile that the new table based overrides are checked for. While some slight overhead  is inevitable for those who make use of the feature (in the order of one or two milliseconds for multiple overrides according to tests on my old laptop), anyone who does not do so will be entirely unaffected. Thus, the primary motivation for the changes was consistency with expected behavior, while a secondary effect was some slight optimizations. A happy side effect is that the config section of each plugin in init.lua actually became more standardized, as configs contained within others.lua were function calls, while ones in their own files simply required the files. Now, nearly everything is a function call.

As a side note for the maintainers, while I normally would have made this all one commit, it was not really possible to do so while maintaining clarity of exactly what was changed/added and why in the commit message as so many files in multiple locations had to be edited to add support. If necessary, I can turn this into one commit, but I think that for house-keeping's sake 3 are necessary.

### DIFF
--- a/lua/plugins/configs/nvimtree.lua
+++ b/lua/plugins/configs/nvimtree.lua
@@ -1,18 +1,20 @@
 local present, nvimtree = pcall(require, "nvim-tree")
-
-local conf = require("core.utils").load_config().plugins.options.nvimtree
-
-local git_status = conf.enable_git
-local ui = conf.ui
-
 if not present then
    return
 end
 
 local g = vim.g
+local default = {
+   conf = require("core.utils").load_config().plugins.options.nvimtree,
+}
+
+default = {
+   git_status = default.conf.enable_git,
+   ui = default.conf.ui,
+}
 
 g.nvim_tree_add_trailing = 0 -- append a trailing slash to folder names
-g.nvim_tree_git_hl = git_status
+g.nvim_tree_git_hl = default.git_status
 g.nvim_tree_highlight_opened_files = 0
 g.nvim_tree_indent_markers = 1
 g.nvim_tree_quit_on_open = 0 -- closes tree when file's opened
@@ -26,7 +28,7 @@ g.nvim_tree_window_picker_exclude = {
 g.nvim_tree_show_icons = {
    folders = 1,
    files = 1,
-   git = git_status,
+   git = default.git_status,
 }
 
 g.nvim_tree_icons = {
@@ -51,7 +53,7 @@ g.nvim_tree_icons = {
    },
 }
 
-local chad_defaults = {
+default = {
    filters = {
       dotfiles = false,
    },
@@ -66,17 +68,18 @@ local chad_defaults = {
       enable = true,
       update_cwd = false,
    },
-   view = ui,
+   view = default.ui,
    git = {
       ignore = false,
    },
 }
 
-M.setup = function (override_flag)
+local M = {}
+M.setup = function(override_flag)
    if override_flag then
-      chad_defaults = require("core.utils").tbl_override_req("nvim_tree", chad_defaults)
+      default = require("core.utils").tbl_override_req("nvim_tree", default)
    end
-   nvimtree.setup(chad_defaults)
+   nvimtree.setup(default)
 end
 
 return M
--- a/lua/plugins/configs/nvimtree.lua
+++ b/lua/plugins/configs/nvimtree.lua
@@ -51,7 +51,7 @@ g.nvim_tree_icons = {
    },
 }
 
-nvimtree.setup {
+local chad_defaults = {
    filters = {
       dotfiles = false,
    },
@@ -71,3 +71,12 @@ nvimtree.setup {
       ignore = false,
    },
 }
+
+M.setup = function (override_flag)
+   if override_flag then
+      chad_defaults = require("core.utils").tbl_override_req("nvim_tree", chad_defaults)
+   end
+   nvimtree.setup(chad_defaults)
+end
+
+return M
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -238,30 +238,32 @@ M.fg_bg = function(group, fgcol, bgcol)
 end
 
 -- Override default config of a plugin based on the path provided in the chadrc
-
--- FUNCTION: override_req, use `chadrc` plugin config override if present
--- name = name inside `default_config` / `chadrc`
--- default_req = run this if 'name' does not exist in `default_config` / `chadrc`
--- if override or default_req start with `(`, then strip that and assume override calls a function, not a whole file
+-- Arguments:
+--   1st - name of plugin
+--   2nd - default config path
+--   3rd - optional function name which will called from default_config path
+--   e.g: if given args - "telescope", "plugins.configs.telescope", "setup"
+--        then return "require('plugins.configs.telescope').setup()"
+--        if 3rd arg not given, then return "require('plugins.configs.telescope')"
 -- if override is a table, mark set the override flag for the default config to true
 -- override flag being true tells the plugin to call tbl_override_req as part of configuration
-M.override_req = function(name, default_req)
-   local override = require("core.utils").load_config().plugins.default_plugin_config_replace[name]
-   local result = default_req
-   if type(override) == "string" then
+M.override_req = function(name, default_config, config_function)
+   local override, apply_table_override =
+      require("core.utils").load_config().plugins.default_plugin_config_replace[name], "false"
+   local result = default_config
+   if type(override) == "string" and override ~= "" then
       result = override
+   elseif type(override) == "table" then
+      apply_table_override = "true"
    end
 
-   if string.match(result, "^%(") then
-      result = result:sub(2)
-      result = result:gsub("%)%.", "').", 1)
-      if type(override) == "table" then
-         result = result:gsub("%(%)", "(true)", 1)
-      end
-      return "require('" .. result
+   result = "('" .. result .. "')"
+   if type(config_function) == "string" and config_function ~= "" then
+      -- add the . to call the functions and concatenate true or false as argument
+      result = result .. "." .. config_function .. "(" .. apply_table_override .. ")"
    end
 
-   return "require('" .. result .. "')"
+   return "require" .. result
 end
 
 -- Override parts of default config of a plugin based on the table provided in the chadrc
@@ -271,7 +273,7 @@ end
 -- default_table = the default configuration table of the plugin
 -- returns the modified configuration table
 M.tbl_override_req = function(name, default_table)
-   local override = require("core.utils").load_config().plugins.default_plugin_config_replace[name]
+   local override = require("core.utils").load_config().plugins.default_plugin_config_replace[name] or {}
    return vim.tbl_deep_extend("force", default_table, override)
 end
 
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -243,21 +243,36 @@ end
 -- name = name inside `default_config` / `chadrc`
 -- default_req = run this if 'name' does not exist in `default_config` / `chadrc`
 -- if override or default_req start with `(`, then strip that and assume override calls a function, not a whole file
+-- if override is a table, mark set the override flag for the default config to true
+-- override flag being true tells the plugin to call tbl_override_req as part of configuration
 M.override_req = function(name, default_req)
    local override = require("core.utils").load_config().plugins.default_plugin_config_replace[name]
    local result = default_req
-
-   if override ~= nil then
+   if type(override) == "string" then
       result = override
    end
 
    if string.match(result, "^%(") then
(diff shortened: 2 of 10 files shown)


Akianonymus:
If i understood it correctly, we cannot override the setup function anymore ?

zbirenbaum:
> If i understood it correctly, we cannot override the setup function anymore ?

@Akianonymus I'm not sure what you mean but I'm pretty sure theres a misunderstanding here.

The behavior if you set the value of any of the override keys in the override default section to a string remains the same. It will still completely ignore the default NvChad configuration. The difference now is, that if you put a table value instead of a string, it will merge that table with the default config overriding values where necessary.

If you take a look in core.utils.override_req(), override ~= nil has instead become a type check. If its nil or any other invalid type it will fail to override and run the default, if it is a string, it returns that string as the config to require same as before, and if it is a table, it modifies the returned require statement such that the configuration knows that there is a table value in chadrc that it needs to merge in. I have around 5 plugins completely overriden so believe me when I say I can gaurantee that the normal override behavior still works perfectly.

Here is the relevant code in core.utils
```
M.override_req = function(name, default_req)
   local override = require("core.utils").load_config().plugins.default_plugin_config_replace[name]
   local result = default_req
   if type(override) == "string" then
      result = override
   end

   if string.match(result, "^%(") then
      result = result:sub(2)
      result = result:gsub("%)%.", "').", 1)
      if type(override) == "table" then
         result = result:gsub("%(%)", "(true)", 1)
      end
      return "require('" .. result
   end

   return "require('" .. result .. "')"
end
```
If I am wrong and you weren't referring to the ability to entirely override a plugin config, please let me know. I didn't include any string values referencing file overrides because it is not part of what I implemented and therefore I thought it may be confusing.

siduck:
@zbirenbaum i will merge this today but before that I will add it in the docs first!

zbirenbaum:
> > If I am wrong and you weren't referring to the ability to entirely override a plugin config, please let me know. I didn't include any string values referencing file overrides because it is not part of what I implemented and therefore I thought it may be confusing.
> 
> I was referring to that, thank you for clarifying, i will test the changes myself to understand it clearly.
> 
> Also, i have made a comment on some lines, reply to that ( reminding in case you didn't see ).
> 
> @siduck do wait for me before merging.

No problem on the clarification, and please do test it. I did my own tests for most of the plugins but it is possible I missed something small. 

Also, could you specify what comment you are referring to? I went through all the files in the 'file changes' area on git but didn't see any comments from review, and the review discussion was empty too. Under reviewers it also says "No reviews" on my end.

lua/plugins/configs/nvimtree.lua
```diff
@@ -71,3 +71,12 @@ nvimtree.setup {
       ignore = false,
    },
 }
+
+M.setup = function (override_flag)
```
Akianonymus, line 75: you missed `local M = {}`

siduck:
> > If I am wrong and you weren't referring to the ability to entirely override a plugin config, please let me know. I didn't include any string values referencing file overrides because it is not part of what I implemented and therefore I thought it may be confusing.
> 
> I was referring to that, thank you for clarifying, i will test the changes myself to understand it clearly.
> 
> Also, i have made a comment on some lines, reply to that ( reminding in case you didn't see ).
> 
> @siduck do wait for me before merging.

alright

Akianonymus:
@siduck  @zbirenbaum Take a look at this, i didn't like how override_req function was working, so refactored it.

https://github.com/Akianonymus/NvChad/commit/ed674b3c70238c26400036b086c816de70e3ebb3

Also handle some other edgecases.

siduck:
@zbirenbaum I request you to make the changes aki added

zbirenbaum:
@Akianonymus Thanks that looks good. I pulled in the changes from that branch to mine and just pushed. Should be ready now @siduck.

zbirenbaum:
Its pretty early here but in a couple of hours I'll try to fix whatever went wrong. I like the style of the changes Aki made but I think theres a bug somewhere. Don't merge this yet @siduck

Akianonymus:
Will look into it.

Akianonymus:
> @Akianonymus Thanks that looks good. I pulled in the changes from that branch to mine and just pushed. Should be ready now @siduck. Edit: wait no something is broken now. I'm confused by your code Aki, where are the tables handled in in it? Also two of my 'string' override plugin configs are broken now

I have pushed a more refactored commit, now working with tables and string both.

https://github.com/Akianonymus/NvChad/commit/babf4c6e984c88d5eec00437ab1c527c3a24511b

Also modified the statusline config to take advantage of the new system.

zbirenbaum:
I pulled in your updates @Akianonymus but while some of the errors seemed to be fixed one remained. I tracked down the issue to be that in your implementation, config_function was still being called in the module referenced by a string in chadrc override

for example, if I had this in my chadrc: 
`signature="custom.plugins.overrides.cmp_configs.lspsignature_cmp",`
because the default function to call for lsp_signature is "signature" in plugins/init.lua, what actually ends up being called for the config function is:
`require("custom.plugins.overrides.cmp_configs.lspsignature_cmp").signature()`
which throws an error for obvious reasons.

I fixed this by changing
```
   if type(override) == "string" and override ~= "" then
      result = override
```
to:
```
   if type(override) == "string" and override ~= "" then
      return "require('" .. override .. "')"
```
So it just returns rather than adding config_function onto the end

I also realized that the way this is currently set up makes it so that a function within a file cannot be called as an override, which is a bit inconvenient. However, the current implementation also makes it very easy to simply return a function to init.lua for the config, which I added support for in override_req

now, the three ways a user can override defaults are:
1. Table
```
nvim_colorizer = { --#000000
   user_default_options = {
      mode = "foreground"
    },
},
```
2. Module path (string)
```
nvim_cmp="custom.plugins.overrides.cmp_configs.cmp", 
```
3. Function
```
feline = function() 
   require("custom.plugins.overrides.statusline_builder.builder").setup()
end,
```

I think this covers all use cases and edge cases so I am happy with the current state if you guys are.

Akianonymus (approved):

Akianonymus:
> .... Aki, good thinking on the statusline refactor, I was thinking of doing that too later on, great feature.

Thank you for tracking the bugs.

siduck:
@Akianonymus cant we import a table from other file in chadrc?

siduck:
@zbirenbaum overriding webdevicons doesnt work for me, does it work there?

Akianonymus:
> @zbirenbaum overriding webdevicons doesnt work for me, does it work there?
> 
> ```
>    nvim_web_devicons = {
>          override = {
>             lua = {
>                icon = "L",
>                name = "lua",
>             },
>          },
>       },
> ```

Did you PackerCompile and restarted nvim ?

zbirenbaum:
> @zbirenbaum overriding webdevicons doesnt work for me, does it work there?
> 
> ```
>    nvim_web_devicons = {
>          override = {
>             lua = {
>                icon = "L",
>                name = "lua",
>             },
>          },
>       },
> ```
This is what appears in my bufferline when I put that code in my config so it's definitely working for me:
2022-01-22-003948_224x55_scrot

siduck:
> nvim_web_devicons = {
>          override = {
>             lua = {
>                icon = "L",
>                name = "lua",
>             },
>          },
>       },

yea works for me when i did packercompile 

https://user-images.githubusercontent.com/59060246/150627053-5f670197-ff24-442c-b8ce-a20312a3cf04.mp4

siduck:
is packercompile a requirement when overriding? i had overriden treesitter,nvimtree config but that didnt require packercompile

zbirenbaum:
> is packercompile a requirement when overriding? i had overriden treesitter,nvimtree config but that didnt require packercompile

It is supposed to be a requirement for consistency and runtime efficiency purposes. 

Can you try editing those plugin overrides again to reproduce that behavior? If it still functions without override I will take a look at it.

siduck:
> > is packercompile a requirement when overriding? i had overriden treesitter,nvimtree config but that didnt require packercompile
> 
> It is supposed to be a requirement for consistency and runtime efficiency purposes.
> 
> Can you try editing those plugin overrides again to reproduce that behavior? If it still functions with
> out override I will take a look at it.

https://user-images.githubusercontent.com/59060246/150627328-d68b7900-d9ab-4d4a-b4d6-04455087e49e.mp4

Akianonymus:
> is packercompile a requirement when overriding? i had overriden treesitter,nvimtree config but that didnt require packercompile

It's a little weird, telescope and statusline also works

> is packercompile a requirement when overriding? i had overriden treesitter,nvimtree config but that didnt require packercompile

You only need to PackerCompile the first time you made changes for a plugin, for example after adding the lua icon changes, try to change it again or change some other language icon, you won't need a PackerCompile.

zbirenbaum:
> > > is packercompile a requirement when overriding? i had overriden treesitter,nvimtree config but that didnt require packercompile
> 
> > 
> 
> > It is supposed to be a requirement for consistency and runtime efficiency purposes.
> 
> > 
> 
> > Can you try editing those plugin overrides again to reproduce that behavior? If it still functions with
> 
> > out override I will take a look at it.
> 
> 
> 
> https://user-images.githubusercontent.com/59060246/150627328-d68b7900-d9ab-4d4a-b4d6-04455087e49e.mp4
> 
> 
> 
> enabling git signs in nvimtree 

Ah I think I understand the confusion now. You didn't change the string nvim tree is set equal to (the "customPlugins.nvimtree") but rather just made changes inside the actual file correct? (If I am wrong please correct me)

Any changes made in that file are meant to take effect during runtime, anything in chadrc's override section requires a recompile.

siduck:
ohh got it :smile:

PR #761, opened 2022-01-29
feat: Add config.plugins.status.snippets to enable/disable snippets

i3Cheese:
I thought many users dont use snippets, so they would like to disable it.

This option remove snippets plugins from order of lazy loading and remove it use from cmp mapping

### DIFF
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -68,6 +68,7 @@ M.plugins = {
       lspsignature = true, -- lsp enhancements
       vim_matchup = true, -- improved matchit
       cmp = true,
+      snippets = true,
       nvimtree = true,
       autopairs = true,
    },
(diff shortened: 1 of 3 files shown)


siduck:
@i3Cheese PR is fine but there's one issue! If the user disables snippets then that would load cmp at start i.e not at insertenter event!

siduck (approved):

siduck:
@i3Cheese completion doesnt work at all :/ 

image

siduck:
image

siduck:
@i3Cheese if cmp makes needs snippets table for sure then ig i have to revert this commit

i3Cheese:
> @i3Cheese if cmp makes needs snippets table for sure then ig i have to revert this commit or you would like to fix this issue

Okay, I will try to do this today.

Issue #765, opened 2022-01-29
Override Entire Packer Use Table


zbirenbaum:
So I think that if nothing else, the whole LSP issue from a bit ago went to show that there is always going to be some niche case in which a user wants to override multiple settings like load order (after, wants, etc) or something else. For instance, I often use forks of plugins that are in the NvChad init.lua for unmerged PRs to test out features or implement my own. 

Also, since I recently did a massive overhaul and reorganization of my LSP and completion setup which improved performance and functionality so much I'll likely PR it soon, which made changes to load order, the use for this type of thing has become pretty apparent to me, as I have to use an edited version of NvChad's init.lua I maintain on my own.

I spent some time thinking about it and came up with a solution that I currently have working but wanted to hear your thoughts regarding this idea @siduck @Akianonymus  before I put too much work into cleaning up parts of the code base it renders unnecessary and organizing.

siduck:
@zbirenbaum yes I was thinking about this too, we can simply use a table instead of function in packer's startup 

see https://github.com/lewis6991/dotfiles/blob/master/config/nvim/lua/lewis6991/plugins.lua#L359 (init is a table in this line)

zbirenbaum:
> @zbirenbaum yes I was thinking about this too, we can simply use a table instead of a function in packer's startup 
> 
> 
> 
> see https://github.com/lewis6991/dotfiles/blob/master/config/nvim/lua/lewis6991/plugins.lua#L359 (init is a table in this line)

Lol that's exactly what I did. I'll go ahead and finish it up today and submit a PR!

siduck:
> > @zbirenbaum yes I was thinking about this too, we can simply use a table instead of a function in packer's startup
> > see https://github.com/lewis6991/dotfiles/blob/master/config/nvim/lua/lewis6991/plugins.lua#L359 (init is a table in this line)
> 
> Lol that's exactly what I did. I'll go ahead and finish it up today and submit a PR!

I guess you need to change stuff for customPlugins too, since user's defined plugins are gonna be in a table too so we need to remove hooks stuff too!

zbirenbaum:
> I guess you need to change stuff for customPlugins too, since user's defined plugins are gonna be in a table too so we need to remove hooks stuff too!

yeah that's what I meant by cleaning up parts of the code base, shouldn't be too hard though.

zbirenbaum:
[...]

PR #767, opened 2022-01-30
Override Entire Packer Use Table

zbirenbaum:
### Implements #765

For the main piece:
Two fields have been added to the default config and are available for use in chadrc:
`default_plugin_override = {}`
and,
`default_plugin_remove = {}`

Their usage is as follows:
default_plugin_override allows individual fields to be overridden and does not affect any fields which the user does not explicitly define like so:
```
default_plugin_override = {
   autopairs = {
      disable = true,    
   },
},
```
default_plugin_remove allows users to entirely remove a plugin from the default plugin table. Unlike "disable" this allows for the user to place the plugin in their own custom.init if they would like. For users which need to override many plugins and options (think lsp issue) using this instead of bloating chadrc with a massive list of overrides is preferable.
```
default_plugin_remove= {
   nvim_tree = true,
   telescope = true, 
},
```

A couple of notes:
First, for clarity purposes I moved the tables for each plugin out into a separate file called plugin_table.lua. This can be moved back into init.lua, but I feel it makes things needlessly unclear to have it there taking up so much space.

Second, while I am aware that the packer startup function can take a table as a parameter rather than iterating over one and running `use()` on each, this would break how we run the user custom.init currently. While I would prefer to have plugin_table and custom.init merged into one table for init to run, I'm not sure this is possible without breaking every user's current custom.init.lua. The prevailing example code for adding a plugin to a config is to put use {...} plugin. While replacing this with a table value and a key would be trivial, it may cause many users some confusion and I wanted to wait for feedback to implement it.

Third, prior override functionality in default_plugin_config_replace has been maintained for now. I can easily remove it, as config_replace's string scenario (replace module) now seems repetitive with features here. The caveat to that is in one specific use case: Overriding individual keys of the configs with tables as added in #735 remains a capability which this change does not replace. I am open to thoughts regarding how this should be approached if changed at all. I was considering only accepting table values in that section so as not to allow for implementing the same functionality in two places, which may cause confusion.


Akianonymus:
@zbirenbaum 

> default_plugin_remove allows users to entirely remove a plugin from the default plugin table. Unlike "disable" this allows for the user to place the plugin in their own custom.init if they would like. For users which need to override many plugins and options (think lsp issue) using this instead of bloating chadrc with a massive list of overrides is preferable.
> 
> ```
> default_plugin_remove= {
>    nvim_tree = true,
>    telescope = true, 
> },
> ```
This can also be achieved by

```lua
default_config_override = {
    nvim_tree = false
    telescope = nil
}
```
Will just remove those plugins as a table merge is happening later.

> While replacing this with a table value and a key would be trivial, it may cause many users some confusion and I wanted to wait for feedback to implement it.
Do elaborate this.

> Overriding individual keys of the configs with tables as added in #735 remains a capability which this change does not replace. ...

Better keep it until a proper alternative is done.

zbirenbaum:
> This can also be achieved by
> 
> ```lua
> default_config_override = {
>     nvim_tree = false
>     telescope = nil
> }
> ```
> 
> Will just remove those plugins as a table merge is happening later.

Could you test that to confirm it? That was my first thought as well but unfortunately vim.tbl_deep_extend threw an error when passed anything other than a table value.

Akianonymus:
> Edit: It was packer that threw the error, if I set it to nil, the key and value aren't even detected in loading the plugin since nil values are treated as non-existent. If I set it to false packer throws an error since it was expecting a table and got a boolean.

Yeah i tested, doesn't work. Also, passing disable = true in packer config isn't equal to removing it ?

which is also done by status.

Akianonymus:
Also, have you tested this ? I am getting a bunch of errors.

zbirenbaum:
> > Edit: It was packer that threw the error, if I set it to nil, the key and value aren't even detected in loading the plugin since nil values are treated as non-existent. If I set it to false packer throws an error since it was expecting a table and got a boolean.
> 
> Yeah i tested, doesn't work. Also, passing disable = true in packer config isn't equal to removing it ?
> 
> which is also done by status

Generally if you try to 'use' the same plugin twice it will throw an error. The idea here is they can remove a plugin from nvchad and configure it in their init.lua. I personally use the status keys to enable and disable my plugins, but because I want the cmp plugin in my personal one enabled does not mean that I also want the one in chad's table enabled.

It's extremely nuanced and niche, but for anyone who likes keeping things extremely organized like me it can make a huge difference...

zbirenbaum:
> Also, have you tested this ? I am getting a bunch of errors.

Extensively

Akianonymus:
> > Also, have you tested this ? I am getting a bunch of errors.
> 
> Extensively, I just swapped from my personal version of NvChad and my personal config to this one. Could you try sync'ing a couple more times and see if that works?

Well, auto pairs gives error on packer compile ( it gets  removed somehow.)

Luasnip gives error too.

Akianonymus:
> It's extremely nuanced and niche, but for anyone who likes keeping things extremely organized like me it can make a huge difference...

Well it's nice and good as long it doesn't break existing stuff.

lua/core/default_config.lua
```diff
@@ -76,20 +76,9 @@ M.plugins = {
       packer = {
          init_file = "plugins.packerInit",
       },
-      autopairs = { loadAfter = "nvim-cmp" },
```
Akianonymus, line 79: why this ?

zbirenbaum: I just realized that autopairs is what is breaking some stuff.  I wanted to remove the "loadafter" stuff since it can be overriden in the new format but there must be something which still references it in the init.lua. I'll fix it real quick.

siduck:
those plugin names arent real so that would cause confusions :/ . I thought you wouldve use just strings like :

 {
      "NvChad/nvim-base16.lua",
      after = "packer.nvim",
      config = function()
         require("colors").init()
      end,
   }

siduck:
I dont think default_plugin_override is needed. We already have default_plugin replace option.

zbirenbaum:
@Akianonymus I just fixed the issue with autopairs and added a check which should fix the other errors.

@siduck 
I really wanted to do that but unfortunately there's no good way to reference them. If someone wanted to override it in that scenario they would have to do it by index in the init file, since it would be a table indexed by number, and that could change after an update.

I tried my best to keep the naming consistent with the "status" section so it shouldn't be too bad.

zbirenbaum:
> I dont think default_plugin_override is needed. We already have default_plugin replace option.

We have default plugin config replace, not default plugin replace. I talked about that above in the third note and Aki commented we should keep it in.

siduck:
I was expecting something like this! 

image

zbirenbaum:
> I was expecting something like this!
> 
> image

How would a user select one of those tables to override though?

siduck:
> > I dont think default_plugin_override is needed. We already have default_plugin replace option.
> 
> We have default plugin config replace, not default plugin replace. I talked about that above in the third note and Aki commented we should keep it in.

btw why default_plugin_override tho? I dont find its use

zbirenbaum:
> > > I dont think default_plugin_override is needed. We already have default_plugin replace option.
> > 
> > 
> > We have default plugin config replace, not default plugin replace. I talked about that above in the third note and Aki commented we should keep it in.
> 
> btw why default_plugin_override tho? I dont find its use

Default plugin override lets a user change a field in the use statement like:
```
feline = {
      "feline-nvim/feline.nvim",
      disable = not plugin_settings.status.feline,
      after = "nvim-web-devicons",
      config = override_req("feline", "plugins.configs.statusline", "setup"),
},
```
```
default_plugin_override = {
      feline = {
           after="something else"
       }
}
```
result:
```
feline = {
      "feline-nvim/feline.nvim",
      disable = not plugin_settings.status.feline,
      after = "something else",
      config = override_req("feline", "plugins.configs.statusline", "setup"),
},
```

zbirenbaum:
Also please let me know before this is merged, filetype and impatient are not in right now because impatient has caused me some strange issues before which causes some things not to throw errors that should until you try it after removing impatient or on a different system.

Whenever it's ready I'll just amend the last commit to include those.

Akianonymus:
I agree with siduck point that they aren't real names so would create problems.

@zbirenbaum a way to do it, although not very clean syntax one can say.

```lua
local plugins = {
 ["feline-nvim/feline.nvim"] = {
         disable = not plugin_settings.status.feline,
          after = "nvim-web-devicons",
          config = override_req("feline", "plugins.configs.statusline", "setup"),
    }
}
```

zbirenbaum:
> I agree with siduck point that they aren't real names so would create problems.
> 
> @zbirenbaum a way to do it, although not very clean syntax one can say.
> 
> ```lua
> local plugins = {
>  ["feline-nvim/feline.nvim"] = {
>          disable = not plugin_settings.status.feline,
>           after = "nvim-web-devicons",
>           config = override_req("feline", "plugins.configs.statusline", "setup"),
>     }
> }
> ```

Ahh the same way lsp handlers are declared. I think that is pretty good actually. I'll go ahead and implement it real quick.

zbirenbaum:
I got it working:

Example config (used disable to easily see if it works by running sync)
```
M.plugins = {
   --default_plugin_config_replace = tbl
   default_plugin_override = { 
      ["kyazdani42/nvim-web-devicons"] = {
         disable = true,
      },
   },
   default_plugin_remove= {
      ["kyazdani42/nvim-tree.lua"] = true,
   },
}
```

siduck:
> I agree with siduck point that they aren't real names so would create problems.
> 
> @zbirenbaum a way to do it, although not very clean syntax one can say.
> 
> ```lua
> local plugins = {
>  ["feline-nvim/feline.nvim"] = {
>          disable = not plugin_settings.status.feline,
>           after = "nvim-web-devicons",
>           config = override_req("feline", "plugins.configs.statusline", "setup"),
>     }
> }
> ```

looks fine but it would be confusing since this doesnt look like packer's syntax :/ . 

wouldve been great if we just used use {...} or just  {...}

zbirenbaum:
> > I agree with siduck point that they aren't real names so would create problems.
> > @zbirenbaum a way to do it, although not very clean syntax one can say.
> > ```lua
> > local plugins = {
> >  ["feline-nvim/feline.nvim"] = {
> >          disable = not plugin_settings.status.feline,
> >           after = "nvim-web-devicons",
> >           config = override_req("feline", "plugins.configs.statusline", "setup"),
> >     }
> > }
> > ```
> 
> looks fine but it would be confusing since this doesnt look like packer's syntax :/ .
> 
> wouldve been great if we just used use {...} or just {...}

It is going to be absolutely necessary to have the brackets in one of the two tables, preferably the one in plugin_table.lua. Would you like me to see if I can make it so that default plugin override can take the form of:
```
default_plugin_override = {
        {
        ...
        }
}
```

siduck:
> > > I agree with siduck point that they aren't real names so would create problems.
> > > @zbirenbaum a way to do it, although not very clean syntax one can say.
> > > ```lua
> > > local plugins = {
> > >  ["feline-nvim/feline.nvim"] = {
> > >          disable = not plugin_settings.status.feline,
> > >           after = "nvim-web-devicons",
> > >           config = override_req("feline", "plugins.configs.statusline", "setup"),
> > >     }
> > > }
> > > ```
> > 
> > 
> > looks fine but it would be confusing since this doesnt look like packer's syntax :/ .
> > wouldve been great if we just used use {...} or just {...}
> 
> It is going to be absolutely necessary to have the brackets in one of the two tables, preferably the one in plugin_table.lua. Would you like me to see if I can make it so that default plugin override can take the form of:
> 
> ```
> default_plugin_override = {
>         {
>         ...
>         }
> }
> ```

yea that looks okay 

for example : 

```
...
  {
      "nvim-telescope/telescope.nvim",
      module = "telescope",
      cmd = "Telescope",
   },
   ...
   
  If the user wants to change something in telescope's defined packer table then 
  
  default_plugin_override = {
        {
          "nvim-telescope/telescope.nvim",
           module = "chad",
        }
}

 
 ```

**Comment by zbirenbaum** [comment]
@siduck Alright, I know how I can do that. Something else to consider is that we could make init.lua take the same format and merge it instead of using packer.run(). Unfortunately, if we do that some things which are set in the default config but not in init would cause unexpected results. It would have to be a merge such that any plugin which is already in the plugin table is replaced, and then all that are not are extended. Also it would break literally every user's config.

Its getting late here though and my girlfriend is apparently unsympathetic to my desire to keep writing code for a text editor so I may have to pick this up in the morning. Let me know your thoughts on putting the overrides into init.lua instead of chadrc. Users would not be able to override by individual keys like they do in default_plugin_override, but it would let them change nvchad's defaults in a more intuitive way.

**Comment by zbirenbaum** [comment]
@siduck 
```
   default_plugin_override = { 
      {
        "akinsho/bufferline.nvim",
         disable = true,
      },
      {
         "nvim-telescope/telescope.nvim",
         disable = true,
      },
   },
```

Now works. All that's left is to decide if you want to keep this in chadrc or make the init.lua changes I mentioned. Then, I'll add filetype and impatient back in, you two can test it, and if all goes well we can merge it.

**Comment by siduck** [comment]
@zbirenbaum we gotta make changes slowly i.e bit by bit. First I'll remove the hooks stuff and clean packer config and then update the docs.

**Comment by siduck** [comment]
@zbirenbaum I request you to re-PR and add only the override , remove stuff. 

I have managed the user table in chadrc and cleaned packer's plugin list
https://github.com/NvChad/NvChad/commit/64c32d2e5e958c70154a4ecd17a771123d0a5662

Also rename override to edit 

```
 default_plugin_edit = {
        {
          "nvim-telescope/telescope.nvim",
           module = "chad",
        }
}
```

zbirenbaum:
@siduck Could we please keep the plugin table in the normal nvchad init.lua in the format I had in this PR? it's not impossible to do it with the format you changed it to, but it does make things much, much harder and messier. The users won't be touching it so it should't matter too much if the format is not exactly what most people expect.

siduck:
@zbirenbaum that format looks scary imo, I want to keep it simple! either use {...} or just {...}. I'll work overrides myself then.

zbirenbaum:
> @zbirenbaum that format looks scary imo, I want to keep it simple! either use {...} or just {...}. I'll work overrides myself then.

I have it working, the bracket format literally just indexes a table by strings

zbirenbaum:
If its not up for discussion to index plugins/init.lua by strings instead of integers i'll finish up documenting and PR it. @siduck

siduck:
> > @zbirenbaum that format looks scary imo, I want to keep it simple! either use {...} or just {...}. I'll work overrides myself then.
> 
> I have it working, the bracket format literally just indexes a table by strings

im already aware that your format works. Its just that the format doesnt look clean imo. Ig its still possible with my format tho. I will see what i can do! Maybe we could just have the default_plugin_remove feature which and not the override one

zbirenbaum:
@siduck It's not that your format doesn't work, its that it adds a for loop of  up to the entire plugin table to every iteration which must be called every time we want to access a specific value instead of just getting it directly.

Saying tbl[x] is a lot easier when debugging or implementing feature than tbl[get_plugin_index(tbl, plugin_name)]

zbirenbaum:
@siduck #780

Issue #774, opened 2022-01-30
Custom plugin 


mikededo:
**Describe the bug**

It does not properly sync if there's only one plugin in the table assigned to the custom props `install` value.

**To Reproduce**

1. Reproduce the installation as explained in the documentation.
2. Copy the `/examples` folder in the `/lua/custom` folder, as explained in the documentation.
3. Modify the `/lua/custom/plugins.lua` file that has been copied from `/examples` as follows
```lua
-- Keep one of the two plugins
return {
  { "elkowar/yuck.vim", ft = "yuck", disable = 1 }
  -- { "ellisonleao/glow.nvim", cmd = "Glow" },
}
```
4. Open `nvim` and run `PackerSync`

**Expected behavior**

It should only install one plugin, the first one, yet it throws the error exposed above. I am no expert with lua and I have not been able to find why could the error be thrown. Is it due to how the prop `install` is parsed? Or it is because of something about `lua` that I am not aware? 😓 

**Screenshots**

No screenshot, but the error recieved is:

```shell
Error detected while processing command line:
E5100: Cannot convert given lua table: table should either have a sequence of positive integer keys or contain only string keys
E5108: Error executing lua error converting argument 1
stack traceback:
        [C]: in function 'expand'
        ...are/nvim/site/pack/packer/opt/packer.nvim/lua/packer.lua:182: in function 'manage'
        ...are/nvim/site/pack/packer/opt/packer.nvim/lua/packer.lua:171: in function 'manage'
        ...are/nvim/site/pack/packer/opt/packer.nvim/lua/packer.lua:320: in function 'manage_all_plugins'
        ...are/nvim/site/pack/packer/opt/packer.nvim/lua/packer.lua:530: in function 'sync'
        [string ":lua"]:1: in main chunk
Press ENTER or type command to continue
```

**Desktop (please complete the following information):**
 - Operating System: Manjaro
 - Terminal: Kitty
 - Version of Neovim: v0.6.1

**Additional context**

No.

Akianonymus:
Try removing the disable flag

mikededo:
It happens with any plugin that is there, as long as there is only one plugin.

Akianonymus:
Will be fixed after the pr is merged..

https://github.com/NvChad/NvChad/pull/776

PR #780, opened 2022-01-31
Packer override

zbirenbaum:
Continuation of #767

### DIFF
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -107,6 +107,8 @@ M.plugins = {
       esc_insertmode_timeout = 300,
    },
    default_plugin_config_replace = {},
+   default_plugin_edit = {},
+   default_plugin_remove = {},
    install = nil,
 }
 
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -107,7 +107,6 @@ M.plugins = {
       esc_insertmode_timeout = 300,
    },
    default_plugin_config_replace = {},
-   default_plugin_edit = {},
    default_plugin_remove = {},
    install = nil,
 }
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -280,4 +280,23 @@ M.tbl_override_req = function(name, default_table)
    return vim.tbl_deep_extend("force", default_table, override)
 end
 
+--Helper function to get key from default plugin table
+local get_plugin_index = function(plugin_table, plugin_name)
+   for index, plugin in ipairs(plugin_table) do
+      if plugin[1] == plugin_name then
+         return index
+      end
+   end
+end
+
+M.remove_default_plugins = function(plugin_table)
+   local removals = require("core.utils").load_config().plugins.default_plugin_remove or {}
+   if vim.tbl_isempty(removals) then return plugin_table end
+   for _, plugin_removal in ipairs(removals) do
+      index = get_plugin_index(plugin_table, plugin_removal)
+      plugin_table[index] = nil
+   end
+   return plugin_table
+end
+
 return M
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -226,8 +226,7 @@ local plugins = {
       end,
    },
 }
---Merge changes to plugin fields from chadrc
-plugins = require("core.utils").edit_default_plugins(plugins)
+--remove plugins specified in chadrc
 plugins = require("core.utils").remove_default_plugins(plugins)
 -- append user plugins to default plugins
 local user_Plugins = plugin_settings.install
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -226,7 +226,9 @@ local plugins = {
       end,
    },
 }
-
+--Merge changes to plugin fields from chadrc
+plugins = require("core.utils").edit_default_plugins(plugins)
+plugins = require("core.utils").remove_default_plugins(plugins)
 -- append user plugins to default plugins
 local user_Plugins = plugin_settings.install
 if type(user_Plugins) == "table" then


siduck:
@zbirenbaum we need to add only the remove feature. It suffices the override feature indirectly too. the override feature's gonna add more confusion imo

zbirenbaum:
@siduck I thought you wanted both? I guess I'll just take out that method then.

siduck:
> @siduck I thought you wanted both? I guess I'll just take out that method then.

No I wanted the remove table only since it can also be used to override

zbirenbaum:
@siduck edits removed

siduck:
great! @zbirenbaum sorry I closed your PR earlier :disappointed:

zbirenbaum:
@siduck Wait I just found a bug that wasn't in the old implementation. It didn't pop up in my first couple tests but for some reason it gliched out just now.

siduck:
btw i tested this rn and it seems to remove my custom plugins too

image

zbirenbaum:
Yeah something is not right

zbirenbaum:
@siduck This is caused by it being indexed by numbers. Apparently the traditional tbl[x] = nil with the format of the current table makes packer error out and get mad, so table.remove has to be used. Unfortunately, table.remove breaks the numbering of the table, so any plugin after the first one is not treated correctly.

siduck:
oh :0

zbirenbaum:
@siduck I pushed a temp fix so that broken code isn't in the repo #782

Unfortunately table.remove is never, ever safe to do in a loop in lua so I didn't do that but found a different way. The code in util is bulky and not very elegant, but it is safe and is necessary until the plugin table is not indexed by integers, so I really, really hope you will reconsider the plugin table format. 

It's your repo so it's whatever you decide of course, but these are the kind of issues I was afraid of when I was being such a pain about the change.

Akianonymus:
> @zbirenbaum we need to add only the remove feature. It suffices the override feature indirectly too. the override feature's gonna add more confusion imo.

How does it suffices the override features ? The goal was to not rewrite every part of the existing packer config and just able to apply incremental changes ( like how it is possible for plugin configs ).

If one removes a plugin, have to write all the config.

Anyway, i am not saying if it should be added or not, just saying that logic is wrong.

zbirenbaum:
> > @zbirenbaum we need to add only the remove feature. It suffices the override feature indirectly too. the override feature's gonna add more confusion imo.
> 
> How does it suffices the override features ? The goal was to not rewrite every part of the existing packer config and just able to apply incremental changes ( like how it is possible for plugin configs ).
> 
> If one removes a plugin, have to write all the config.
> 
> Anyway, i am not saying if it should be added or not, just saying that logic is wrong.

This was my understanding as well

siduck:
> > @zbirenbaum we need to add only the remove feature. It suffices the override feature indirectly too. the override feature's gonna add more confusion imo.
> 
> How does it suffices the override features ? The goal was to not rewrite every part of the existing packer config and just able to apply incremental changes ( like how it is possible for plugin configs ).
> 
> If one removes a plugin, have to write all the config.
> 
> Anyway, i am not saying if it should be added or not, just saying that logic is wrong.

override lets us change change things in a plugin's defined table {..stuff here..}

But with the remove table we can do the same but extra work. 

remove_table = {"plugin name"}

and then in your custom plugin list 

re-define that plugin in your custom plugins file with your desired change

Akianonymus:
Nevermind 😂.

siduck:
> @siduck I pushed a temp fix so that broken code isn't in the repo #782
> 
> Unfortunately table.remove is never, ever safe to do in a loop in lua so I didn't do that but found a different way. The code in util is bulky and not very elegant, but it is safe and is necessary until the plugin table is not indexed by integers, so I really, really hope you will reconsider the plugin table format.
> 
> I don't want to sound like a broken record and it's your repo so it's whatever you decide of course, but these are the kind of issues I was afraid of when I was being such a pain about the change.

idk much lua but how's table.remove not safe? Ive tried deleting a table by setting nil to it but that doesnt work completely. It removes the last table (in plugins table) so like telescope was defined at the last in our plugins table. For example if I added nvimtree in the remove table then the function removes nvimtree+telescope which is very weird. But it worked perfectly wit table.remove

siduck:
https://user-images.githubusercontent.com/59060246/151748491-b3b36946-6a03-44fe-9563-f35389c34518.mp4

siduck:
lets make this change :laughing: :rocket:

zbirenbaum:
> lets make this change :laughing: :rocket: 

Please don't

siduck:
oh no :c

siduck:
the code looked very clean and it works perfect tho!

zbirenbaum:
> the code looked very clean and it works perfect tho!

If you look up table.remove there are tons of threads on why you shouldn't use it.

The long story short is that table remove shifts the indices when it removes a value which can lead to some extraordinarily unexpected behavior inside loops. 

In this particular situation it may work but try it with multiple plugins beide you do anything because that's where the issues arise.

In general it's almost universally considered awful practice to use inside of loops like this even if it works in our scenario, if for no other reason than if changed even slightly it can introduce tons of bugs and using it safely is requires profoundly inefficient code.

siduck:
> > @siduck I pushed a temp fix so that broken code isn't in the repo #782
> > Unfortunately table.remove is never, ever safe to do in a loop in lua so I didn't do that but found a different way. The code in util is bulky and not very elegant, but it is safe and is necessary until the plugin table is not indexed by integers, so I really, really hope you will reconsider the plugin table format.
> > I don't want to sound like a broken record and it's your repo so it's whatever you decide of course, but these are the kind of issues I was afraid of when I was being such a pain about the change.
> 
> idk much lua but how's table.remove not safe? Ive tried deleting a table by setting nil to it but that doesnt work completely. It removes the user's table + remove plugin table's plugins which is very weird. But it worked perfectly with table.remove

@Akianonymus should i make the change?

Akianonymus:
Wait, i will make a new pr.

siduck:
i hope its not regarding changing the format! If we cant use table.remove then table.insert can be used 

image

Akianonymus:
> i hope its not regarding changing the format! If we cant use table.remove then table.insert can be used
> 
> image

This looks fine to me

zbirenbaum:
> i hope its not regarding changing the format! If we cant use table.remove then table.insert can be used 
> 
> 
> 
> image
> 
> 

This looks perfect, thank you for not using table.remove!

siduck:
> > i hope its not regarding changing the format! If we cant use table.remove then table.insert can be used
> > image
> 
> This looks perfect, thank you for not using table.remove!

table.remove seemed cleaner tho! ( I didnt had to make a new table) . Can you share why was it unsafe? my mind's still tinkering to add it :rocket: :smiley:

zbirenbaum:
> > > i hope its not regarding changing the format! If we cant use table.remove then table.insert can be used
> > > image
> > 
> > 
> > This looks perfect, thank you for not using table.remove!
> 
> table.remove seemed cleaner tho! (I didnt had to make a new table) . Can you share why was it unsafe? my mind's still tinkering to add it rocket smiley

Did you read any of the massive essay I wrote above along with any of the posts in the link I included in it?

By the way I can't get custom plugins working even with a minimal chadrc and  table right now. Tons of errors and eventually packer just starts uninstalling itself.

Honestly just add it, I'm probably just going to have to maintain my own fork with plugins indexed properly so this doesn't keep acting up while I have important stuff to do like rn.

Akianonymus:
@zbirenbaum what errors ? It's working fine here.

zbirenbaum:
> @zbirenbaum what errors ? It's working fine here.

Same one as #774 but with one or multiple plugins @Akianonymus
I cloned a fresh repo and tried to make a custom plugins with just trouble.nvim and it breaks down and doesn't recover the second I add the plugin. Even removing the plugin and all references to the custom table won't fix it, I have to delete the packer_compiled file.

Akianonymus:
Weird, working fine here.

zbirenbaum:
> Weird, working fine here.

@Akianonymus The error sometimes backtraces to something in utils.options. I think some setting in options is broken now, but it's not the shadafile as I tried commenting that, and idk what else changed. I honestly don't know whats going on but I'm very frustrated, as I have around a thousand lines of CUDA C I need to write, and don't have time to spend an hour tracking down whatever it is, but maybe I can in a couple days.

All I know is, when I make everything the format that I had before, literally just these lines are necessary to make everything work flawlessly, so I am very, very frustrated that things need to be so needlessly complex for "source aesthetic" purposes:

```
local custom_table = require("custom.plugins_table")
return packer.startup(function()
   for k,v in pairs(custom_table) do
      plugins[k] = v
   end
   for _,v in pairs(plugins) do
      use(v)
   end
end)
```
With that block and string indexing I have none of the issues described above so for now I'm going to stay on my stable branch until I can see if it can be fixed.

Akianonymus:
Well, provide a log or something, or steps to reproduce, everything working fine here.

Link your custom changes too.

siduck:
> Well, provide a log or something, or steps to reproduce, everything working fine here.
> 
> Link your custom changes too.

path of table works but not a table within the install variabl 

> > > > i hope its not regarding changing the format! If we cant use table.remove then table.insert can be used
> > > > image
> > > 
> > > 
> > > This looks perfect, thank you for not using table.remove!
> > 
> > 
> > table.remove seemed cleaner tho! (I didnt had to make a new table) . Can you share why was it unsafe? my mind's still tinkering to add it rocket smiley
> 
> Did you read any of the massive essay I wrote above along with any of the posts in the link I included in it?
> 
> By the way I can't get custom plugins working even with a minimal chadrc and table right now. Tons of errors and eventually packer just starts uninstalling itself.
> 
> Honestly just add it, I'm probably just going to have to maintain my own fork with plugins indexed properly so this doesn't keep acting up while I have important stuff to do like rn.

my bad i missed that essay

siduck:
https://user-images.githubusercontent.com/59060246/151926858-78cc0b6f-7448-417d-b0fd-8a978bac36e2.mp4

works fine here

siduck:
https://user-images.githubusercontent.com/59060246/151926858-78cc0b6f-7448-417d-b0fd-8a978bac36e2.mp4

works fine here 

make sure you remove this var if you dont use it at all 
image

zbirenbaum:
@Akianonymus Sorry I got quite busy for a bit. I did some looking into it, and I think there was something on my end left over from a different edition of the implementation causing the issue, especially since no other users report a problem. For now though, I like the format and simplicity of my setup and merging in changes from NvChad's main repo isn't too hard, so I think I'll just use my version for personal development and maintain a fairly minimal custom dir and vanilla branch for NvChad development, so don't worry about it too much.

PR #783, opened 2022-01-31
improve startup time with shada load deferral

zbirenbaum:
This just sets shada to none so it isn't read during startup, and then sets it to the normal location and reads it after startup. For some users the difference may be negligible, but many plugins will write to shada and over time command history etc fills up and this can save a lot of time.

Once, before discovering this, I saw shada took like 15ms to load and I checked it, and it had over 13k lines.... I had to just delete the file to restore startup times, this should avoid situations like that.

Even having cleaned it pretty recently, this change saves me around 3-4ms for each startup.

### DIFF
--- a/lua/core/options.lua
+++ b/lua/core/options.lua
@@ -3,6 +3,13 @@ local g = vim.g
 
 local options = require("core.utils").load_config().options
 
+--Defer loading shada until after startup
+vim.opt.shadafile = "NONE"
+vim.defer_fn(function()
+   vim.opt.shadafile = vim.fn.expand("$HOME") .. "/.local/share/nvim/shada/main.shada"
+   vim.cmd[[ rsh ]]
+end, 0)
+
 opt.title = true
 opt.clipboard = options.clipboard
 opt.cmdheight = options.cmdheight


siduck (approved):

Akianonymus:
What if user had a custom path set for shada ? Also as discussed previously, vim.defer_fn still runs on the main ui thread, blocking anything before doing any stuff, or is this not true ?

zbirenbaum:
> What if user had a custom path set for shada ? Also as discussed previously, vim.defer_fn still runs on the main ui thread, blocking anything before doing any stuff, or is this not true ?


I'll add a variable to set Shada path. When I don't use defer startuptime reports about 3ms for reading shada. When I do use it, its reported as .003 or something equivalent. I believe defer_fn implicitly calls schedule_wrap which should wait until after main loop but I may be wrong.

zbirenbaum:
> What if user had a custom path set for shada ? Also as discussed previously, vim.defer_fn still runs on the main ui thread, blocking anything before doing any stuff, or is this not true ?

Here's the source for defer_fn: 
```
function vim.defer_fn(fn, timeout)
  vim.validate { fn = { fn, 'c', true}; }
  local timer = vim.loop.new_timer()
  timer:start(timeout, 0, vim.schedule_wrap(function()
    timer:stop()
    timer:close()

    fn()
  end))

  return timer
end
```

Calling it with 0 is essentially equal to a schedule wrap call. I checked and schedule wrap offers the same startuptime improvement. I'll change it to use schedule wrap to avoid confusion when I add in the var though.

PR #784, opened 2022-01-31
restore shada location before load

zbirenbaum:
Also changed defer to schedule. Same effect but clearer what happens I guess.

### DIFF
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -26,6 +26,7 @@ M.options = {
    updatetime = 250,
    undofile = true,
    fillchars = { eob = " " },
+   shadafile = vim.opt.shadafile,
 
    -- NvChad options
    nvChad = {
--- a/lua/core/options.lua
+++ b/lua/core/options.lua
@@ -73,3 +73,10 @@ local disabled_built_ins = {
 for _, plugin in pairs(disabled_built_ins) do
    g["loaded_" .. plugin] = 1
 end
+
+--Defer loading shada until after startup_
+vim.opt.shadafile = "NONE"
+vim.schedule(function()
+   vim.opt.shadafile = require("core.utils").load_config().options.shadafile
+   vim.cmd [[ rsh ]]
+end)


zbirenbaum:
@Akianonymus This fix the issues?

Akianonymus:
> @Akianonymus This fix the issues?

Still hardcoding the shada path, just fetch the initial shada file name, store it in a var, then restore after the defer_fn and stuff.

also, vim.opt.shadafile is a table, idk how that is gonna work.

Akianonymus:
Also, the main point being, does this even save some time ? 🤔

Given shada file perfoms some critical tasks.

```
The ShaDa file is used to store:
  - The command line history.
  - The search string history.
  - The input-line history.
  - Contents of non-empty registers.
  - Marks for several files.
  - File marks, pointing to locations in files.                         
  - Last search/substitute pattern (for 'n' and '&').
  - The buffer list.                                                    
  - Global variables.
```

zbirenbaum:
> > @Akianonymus This fix the issues?
> 
> Still hardcoding the shada path, just fetch the initial shada file name, store it in a var, then restore after the defer_fn and stuff.
> 
> also, vim.opt.shadafile is a table, idk how that is gonna work.

It worked perfectly when I tested it. I have always set shada with a string, try it out for yourself though. I just amended

zbirenbaum:
> Also, the main point being, does this even save some time ? thinking
> 
> Given shada file perfoms some critical tasks.
> 
> ```
> The ShaDa file is used to store:
>   - The command line history.
>   - The search string history.
>   - The input-line history.
>   - Contents of non-empty registers.
>   - Marks for several files.
>   - File marks, pointing to locations in files.                         
>   - Last search/substitute pattern (for 'n' and '&').
>   - The buffer list.                                                    
>   - Global variables.
> ```

Once it is read all of that is available as normal. Pull the changes and press up in your command history and you should get the last command you ran despite it being loaded later. Shada gets really, really full and when I don't do this or completely delete it for a while has historically been one of the largest components of my startup time.

Akianonymus:
> > > @Akianonymus This fix the issues?
> > 
> > 
> > Still hardcoding the shada path, just fetch the initial shada file name, store it in a var, then restore after the defer_fn and stuff.
> > also, vim.opt.shadafile is a table, idk how that is gonna work.
> 
> It worked perfectly when I tested it. I have always set shada with a string, try it out for yourself though. I just amended

I was confused because vim.opt.shadafile is a table, i tried to print it.

Akianonymus:
> > Also, the main point being, does this even save some time ? thinking
> > Given shada file perfoms some critical tasks.
> > ```
> > The ShaDa file is used to store:
> >   - The command line history.
> >   - The search string history.
> >   - The input-line history.
> >   - Contents of non-empty registers.
> >   - Marks for several files.
> >   - File marks, pointing to locations in files.                         
> >   - Last search/substitute pattern (for 'n' and '&').
> >   - The buffer list.                                                    
> >   - Global variables.
> > ```
> 
> Once it is read all of that is available as normal. Pull the changes and press up in your command history and you should get the last command you ran despite it being loaded later. Shada gets really, really full and when I don't do this or completely delete it for a while has historically been one of the largest components of my startup time.

how do you measure the startup time ? Lemme remind you vim default startuptime is very useless.

---

https://github.com/Akianonymus/NvChad/commit/88b0a2de78950f1f25ab1492c2b9b810bba5c566

pick this commit

Also, squash all the other commits with this, no need to author me.

zbirenbaum:
> > Once it is read all of that is available as normal. Pull the changes and press up in your command history and you should get the last command you ran despite it being loaded later. Shada gets really, really full and when I don't do this or completely delete it for a while has historically been one of the largest components of my startup time.
> 
> how do you measure the startup time ? Lemme remind you vim default startuptime is very useless.
> 
> [Akianonymus@88b0a2d](https://github.com/Akianonymus/NvChad/commit/88b0a2de78950f1f25ab1492c2b9b810bba5c566)
> 
> pick this commit
> 
> Also, squash all the other commits with this, no need to author me.

Thanks, I think I got it committed properly. 

I generally check startup time with a mix of startuptime and packer profile. Neither of those are particularly useful in this case. What I will say is that shada can have a monumental impact on startup time. A while ago my shada was absurdly long, and I found out while trying to discover why my startup time had noticeably slowed. 

Theoretically, this should be a fix as the scheduled function should fire after the event loop, but if not then there's no real harm in the change I suppose. 

I suppose the way to tell would be to have a 10-20k line shada like my old one and see if it noticeably speeds things up.

I found someone else that agrees with me here with a quick google search, although reddit isn't the best source, and he said it didn't make much of a difference for him. I think NvChad uses some plugins that write to shada though considering how long mine gets so it may help some of our users more: 
https://www.reddit.com/r/neovim/comments/opipij/guide_tips_and_tricks_to_reduce_startup_and/
Disable the shadafile while sourcing config
This didn't make much of a difference, but apparently its good behavior

vim.opt.shadafile = "NONE" 
vim.opt.shadafile = ""
The former goes under init.lua, and the latter goes at the end of util.lua

Akianonymus:
https://youtu.be/GToxQKi7LZg

Should read this.

Also, pr ready to merge.

PR #785, opened 2022-01-31
Remove duplicate code, missed in d8dabe9 | Format files

Akianonymus:


Akianonymus:
@zbirenbaum @siduck

You both often not format the files, try to do, in a PR or when pushed directly ( siduck ).

It's a simple command, `stylua .`, will format all the files.

siduck (approved):

zbirenbaum:
> @zbirenbaum @siduck
> 
> 
> 
> You both often not format the files, try to do, in a PR or when pushed directly ( siduck ).
> 
> 
> 
> It's a simple command, `stylua .`, will format all the files.

Oops sorry about that. I have my personal config set up to match the format of NvChad's source but it gets messed up sometimes when I use a clean config for testing. I'll use the stylua in the future tho.

On that note, it may be a good idea to have the defaults match the formatting of the codebase. It's seems a bit counter intuitive that the two don't match.

PR #786, opened 2022-01-31
utils: Fix removal of plugins with chadrc

Akianonymus:
@siduck @zbirenbaum check if this works.

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -285,21 +285,14 @@ M.remove_default_plugins = function(plugin_table)
    if vim.tbl_isempty(removals) then
       return plugin_table
    end
-   local newtable = {}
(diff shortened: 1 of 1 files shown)


siduck:
@Akianonymus i tried this already. It works but packer will give error on empty tables defiined.

image

So only thing we can do is use table.remove or table.insert as i mentioned earlier which works 100%

Akianonymus:
> @Akianonymus i tried this already. It works but packer will give error on empty tables defiined.
> 
> image
> 
> So only thing we can do is use table.remove or table.insert as i mentioned earlier which works 100%

Ooh i failed to see.

Also where is your code, i would like to see.

Akianonymus:
@siduck insert one is good.

PR #796, opened 2022-02-03
prioritize themes outside of nvim-base16.lua

max397574:


siduck:
Will be added in the future, not this soon!

PR #811, opened 2022-02-10
Terminal toggle

zbirenbaum:
This is a draft/RFC and not remotely ready to be merged.

Core functionality is all fine, <leader>h and <leader>v will spawn their respective terminals and hide them if within the terminal. Since hiding is what JK did before this hotkey is now useless and will be removed, especially since it does not close the window spawned in split leading to visual glitches.

Telescope picks up the terminals in the buffer table, but <leader>W will not register them. Not sure why this is, but it can definitely be fixed since telescope does detect the terminal buffer when hidden.

Options have been added to default_config and chadrc for customizing the height and width of the splits.

spawning a terminal in a new window has not yet been implemented, but should be trivial. Honestly since the new window isn't really "togglable" beyond opening a different buffer, we can probably just keep the old implementation.


zbirenbaum:
> I thought this was gonna go in the extensions repo!

I can move some of it there. A lot of the Whitney stuff is set here.

I haven't had much time to work on this lately. For now I would add a flag in the config which allows people to disable the terminal stuff in autocmds which is what breaks toggleterm and other terminal plugins.

I can move this to the extensions repo and work on it when I have time there.

siduck:
@zbirenbaum can you add this in the extensions repo?

zbirenbaum:
@siduck I moved the core terminal implementation to NvChad/extensions#10. This PR now only contains the strictly necessary changes to lua/plugins/init.lua, mappings.lua, and default_config.lua. All previous commits related to the terminal have been moved to the NvChad/extensions PR. Everything seems to be working fine with it on my end, so to test it out you can use this PR and then just packer use zbirenbaum/extensions instead of NvChad/extensions in the init.lua

I think this is ready to be merged from my testing.

evprkr:
When will this be merged? Just set up NvChad today and I'm looking to get this functionality but don't want to clone this PR if it's going to be merged in a few days.

siduck:
@evprkr do packersync

zbirenbaum:
> @evprkr do packersync 

Yeah that was gonna be my advice

evprkr:
> @evprkr do packersync

Just did this, still having some weird issues. The first time I do `<leader>h`, I get a terminal in a new tab. If I then run 'exit' in that terminal to go back to where I was at, I then have a split at the bottom of my buffer with what appears to be a copy of my buffer. I can `:q` that, then doing `<leader>h` again opens a terminal like I'd expect it to.

However, then entering `JK` gives me this error:
```
E5108: Error executing lua [string ":lua"]:1: attempt to call field 'hide' (a nil value)
stack traceback:
        [string ":lua"]:1: in main chunk
Press ENTER or type command to continue
```

My config is almost completely stock, I just have a few things like tabstop, shiftwidth, theme, etc, and for plugin configs, I just have some cosmetic stuff for bufferline.

zbirenbaum:
@evprkr I just tested it, I can say with certainty that the issue is you are running the old mappings somehow, as JK now shouldn't do anything except go to normal mode inside the buffer. making JK hide it is something I forgot to implement to match up with the previous implementation. It would  require a little more logic, so I'll implement it later today and PR the extensions repo.

evprkr:
> @evprkr ~I just tested it, I can say with certainty that the issue is you are running the old mappings somehow, as JK now shouldn't do anything except go to normal mode inside the buffer. making JK hide it is something I forgot to implement to match up with the previous implementation. It would require a little more logic, so I'll implement it later today and PR the extensions repo.~
> 
> edit: This line is wrong and may be the cause:
> 
> ```
> map("t", term_maps.esc_hide_termmode, "<CMD>lua require('nvchad.terminal').hide() <CR>")
> ```
> 
> I'll have it changed. Try deleting it and see if it fixes your issue, but on a clean install it does nothing for me, which is probably why I didn't notice sooner.
> 
> With the defaults, spacebar+h is all you should use to show/hide horizontal, and spacebar+v vertical

I'm assuming you found that from the gists I posted in a separate issue just a bit ago, links here in case anyone else wants them:

https://gist.github.com/evprkr/030528c3a983f57625eceff0a74e246f
https://gist.github.com/evprkr/eaff10f91d8199b04cdcbd883c3fe48b

That line wasn't actually added until after I posted my issue here. Is there anything else weird? I'm removing that line anyways.

zbirenbaum:
> > @evprkr ~I just tested it, I can say with certainty that the issue is you are running the old mappings somehow, as JK now shouldn't do anything except go to normal mode inside the buffer. making JK hide it is something I forgot to implement to match up with the previous implementation. It would require a little more logic, so I'll implement it later today and PR the extensions repo.~
> > edit: This line is wrong and may be the cause:
> > ```
> > map("t", term_maps.esc_hide_termmode, "<CMD>lua require('nvchad.terminal').hide() <CR>")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'll have it changed. Try deleting it and see if it fixes your issue, but on a clean install it does nothing for me, which is probably why I didn't notice sooner.
> > With the defaults, spacebar+h is all you should use to show/hide horizontal, and spacebar+v vertical
> 
> I'm assuming you found that from the gists I posted in a separate issue just a bit ago, links here in case anyone else wants them:
> 
> https://gist.github.com/evprkr/030528c3a983f57625eceff0a74e246f https://gist.github.com/evprkr/eaff10f91d8199b04cdcbd883c3fe48b
> 
> That line wasn't actually added until after I posted my issue here. Is there anything else weird? I'm removing that line anyways. Edit: Just confirmed again, removing the line does not change the terminal behavior.
> 
> Is there a way I can reset everything outside of these two files? Since I spent all day yesterday trying to figure out Neovim (coming from Vim was harder than I thought) it's possible I changed something else without realizing.

I'm a little confused here, the line I referenced was inside lua/core/mappings. If it was added after you posted @siduck may have commit amend'd?

Any code you write for the configuration should ONLY be inside of the lua/custom directory. I made a basic custom dir with the two files you posted and didn't run into any issues. I think you should try to git pull inside of NvChad or just move your custom dir outside of it, nuke, reclone,  move your custom dir into the fresh clone, and packer sync.

evprkr:
@zbirenbaum Ah I think I misunderstood. I thought he was referencing a line I had in my customs that is sort of similar. I'll try what he said (deleting that line) I'll try that, and if that doesn't work I'll try nuking the whole thing, and report back.

evprkr:
> @evprkr ~I just tested it, I can say with certainty that the issue is you are running the old mappings somehow, as JK now shouldn't do anything except go to normal mode inside the buffer. making JK hide it is something I forgot to implement to match up with the previous implementation. It would require a little more logic, so I'll implement it later today and PR the extensions repo.~
> 
> edit: This line is wrong and may be the cause:
> 
> ```
> map("t", term_maps.esc_hide_termmode, "<CMD>lua require('nvchad.terminal').hide() <CR>")
> ```
> 
> I'll have it changed. Try deleting it and see if it fixes your issue, but on a clean install it does nothing for me, which is probably why I didn't notice sooner.
> 
> With the defaults, spacebar+h is all you should use to show/hide horizontal, and spacebar+v vertical

I can confirm that commenting out this line fixes the issue! Was that line still supposed to be there after updating and syncing, or did I do something weird?

zbirenbaum:
> > @evprkr ~I just tested it, I can say with certainty that the issue is you are running the old mappings somehow, as JK now shouldn't do anything except go to normal mode inside the buffer. making JK hide it is something I forgot to implement to match up with the previous implementation. It would require a little more logic, so I'll implement it later today and PR the extensions repo.~
> 
> > 
> 
> > edit: This line is wrong and may be the cause:
> 
> > 
> 
> > ```
> 
> > map("t", term_maps.esc_hide_termmode, "<CMD>lua require('nvchad.terminal').hide() <CR>")
> 
> > ```
> 
> > 
> 
> > I'll have it changed. Try deleting it and see if it fixes your issue, but on a clean install it does nothing for me, which is probably why I didn't notice sooner.
> 
> > 
> 
> > With the defaults, spacebar+h is all you should use to show/hide horizontal, and spacebar+v vertical
> 
> 
> 
> I can confirm that commenting out this line fixes the issue! Was that line still supposed to be there after updating and syncing, or did I do something weird?

It's my mistake I think. That was part of the old implementation but currently there's no direct translation for the new terminals. Showing and hiding necessarily takes direction as a parameter. I may be able to restore the functionality since if the user is in terminal mode we know the current buffer is a terminal. I'll see what I can do about adding a directionless hide in, but in the meantime that code needs to be commented or removed @siduck

evprkr:
> > > @evprkr ~I just tested it, I can say with certainty that the issue is you are running the old mappings somehow, as JK now shouldn't do anything except go to normal mode inside the buffer. making JK hide it is something I forgot to implement to match up with the previous implementation. It would require a little more logic, so I'll implement it later today and PR the extensions repo.~
> > 
> > 
> > > 
> > 
> > 
> > > edit: This line is wrong and may be the cause:
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > map("t", term_maps.esc_hide_termmode, "lua require('nvchad.terminal').hide() ")
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > I'll have it changed. Try deleting it and see if it fixes your issue, but on a clean install it does nothing for me, which is probably why I didn't notice sooner.
> > 
> > 
> > > 
> > 
> > 
> > > With the defaults, spacebar+h is all you should use to show/hide horizontal, and spacebar+v vertical
> > 
> > 
> > I can confirm that commenting out this line fixes the issue! Was that line still supposed to be there after updating and syncing, or did I do something weird?
> 
> It's my mistake I think. That was part of the old implementation but currently there's no direct translation for the new terminals. Showing and hiding necessarily takes direction as a parameter. I may be able to restore the functionality since if the user is in terminal mode we know the current buffer is a terminal. I'll see what I can do about adding a directionless hide in, but in the meantime that code needs to be commented or removed @siduck

Ah okay. As it turns out I was mistaken, the issue is not fixed by removing that line. The issue I'm having only happens when I have more than one tab open (ie. through `vim -p file.lua file2.lua` or something else)

I'm also noticing that my bufferline configs aren't even working. They were before I updated, but not anymore. Something broke. I guess I'm going to just backup my customs and nuke the whole thing.

evprkr:
Okay, something is going on here and I'm not sure what. Completely removed `~/.config/nvim` ran the install from the wiki again, did the sync, restored my customs, still having the same issue, bufferline configs are still not working.

zbirenbaum:
> Okay, something is going on here and I'm not sure what. Completely removed `~/.config/nvim` ran the install from the wiki again, did the sync, restored my customs, still having the same issue, bufferline configs are still not working.

I don't touch bufferline, I have the plugin removed. I think it was updated recently so you may want to make a separate issue for it and ask @siduck

evprkr:
Good idea. Thanks.

PR #818, opened 2022-02-13
improve method of adding and removing plugins

zbirenbaum:
@siduck @Akianonymus  This is a solution which I should have thought of ages ago during the dispute around having a labeled plugin table or not. This PR leaves the current table unchanged and therefore "pretty" as it is, but the functions for merging changes from chadrc are vastly simplified.

I've been having issues with an error message about using the same plugins multiple times on vanilla NvChad with just a basic chadrc, and the current logic for adding and removing plugins, while improved since the first iteration, remains somewhat difficult to debug and therefore is not ideal. These changes fix all that.

Notable improvements include:
1. If the user has a plugin in their custom plugin table which is already in the default table, their plugin will automatically take the place of the default one without error.
2. There are never any errors or edge cases.
3. The user can have a labeled or unlabeled plugin table and it will accept either.
4. By extension of 3, it's not a breaking change
5. The code for adding and removing plugins is much, much cleaner
6. The code for adding and removing plugins is much, much more efficient
7. I don't have to maintain a fork anymore lmao

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -280,23 +280,35 @@ M.tbl_override_req = function(name, default_table)
    return vim.tbl_deep_extend("force", default_table, override)
 end
 
-M.remove_default_plugins = function(plugin_table)
-   local removals = require("core.utils").load_config().plugins.default_plugin_remove or {}
-   local result = {}
-
-   if vim.tbl_isempty(removals) then
-      return plugin_table
+--provide labels to plugins instead of integers
+M.label_plugins = function(plugins)
+   plugins_labeled = {}
+   for _, plugin in ipairs(plugins) do
+      plugins_labeled[plugin[1]] = plugin
    end
+   return plugins_labeled
+end
 
-   for _, value in pairs(plugin_table) do
-      for _, plugin in ipairs(removals) do
-         if value[1] ~= plugin then
-            table.insert(result, value)
-         end
+-- remove plugins specified by user from the plugins table
+M.remove_default_plugins = function(plugins)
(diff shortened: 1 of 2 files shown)


siduck:
Our current code looks simpler than this :[

siduck:
I'm on phone rn and the code looks intimidating for a lua noob as me, if we were to merge this then we will have to remove these 2 functions as well 

Screenshot_20220213-143551_DuckDuckGo
 
Since the user can replace the default config of a plug-in as mentioned in (1)

zbirenbaum:
> I'm on phone rn and the code looks intimidating for a lua noob as me, if we were to merge this then we will have to remove these 2 functions as well 
> 
> 
> 
> Screenshot_20220213-143551_DuckDuckGo
> 
>  
> 
> Since the user can replace the default config of a plug-in as mentioned in (1)

Our current code is much more complex than this, I really don't know how much simpler it can get than iterating over a table in a loop... all the init.lua code does is make each index of the table have the label "Someuser/somerepo". 

To be honest if you are a "lua noob" I'm really going to ask you to trust me on this one, as ever since you changed a ton in that last PR I've had tons of issues and can't use vanilla NvChad. All this code follows very standard conventions so I'm really not sure what you are talking about. 

Default config override doesn't need to be changed, it provides different functionality and only changes the config file. If it would have to be removed from this, it shouldn't be in currently either.

siduck (approved):

Akianonymus:
Nice to see this was merged.

Btw, string catenation is intimidating, xD.

Issue #829, opened 2022-02-18
`table index is nil` when doing `PackerSync`


sitiom:
**Describe the bug**
I get the error `E5108: Error executing lua C:\Users\rycae\AppData\Local\nvim/lua/core/utils.lua:308: table index is nil` when doing PackerSync.

This bug was introduced in https://github.com/NvChad/NvChad/pull/818. For now, I will be reverting to the last working commit (https://github.com/NvChad/NvChad/commit/a21e884983ef6d99ef6fb3d3ce714953c43332a1).

**To Reproduce**
1. Do `:PackerSync`

**Expected behavior**
My config should work fine as before.

**Screenshots**
[![asciicast](https://asciinema.org/a/ZCNaQS8YUSQWuUGrjeD6J0Pjq.svg)](https://asciinema.org/a/ZCNaQS8YUSQWuUGrjeD6J0Pjq)

**Desktop (please complete the following information):**
 - Windows 11
 - Windows Terminal
 - Neovim v0.6.1

**Additional context**
This is not a Windows-specific error.

siduck:
it works fine here

https://user-images.githubusercontent.com/59060246/154801843-22f4bff2-0aa3-4e13-a158-d8ce9ab41564.mp4

looks like some windows/packer issue :l

sitiom:
It is not a windows issue because it also happens to me on Linux. It might not be also a packer issue because I am on https://github.com/NvChad/NvChad/commit/a21e884983ef6d99ef6fb3d3ce714953c43332a1 and the latest packer commit is working fine.

siduck:
> It is not a windows issue because it also happens to me on Linux. It might not be also a packer issue because I am on [a21e884](https://github.com/NvChad/NvChad/commit/a21e884983ef6d99ef6fb3d3ce714953c43332a1) and the latest packer commit is working fine.

can you update to the latest commit?

sitiom:
This bug happens on the latest commit of nvchad.

sitiom:
Additional context: it works fine with an empty user plugins table (`return { }`). As soon as I add one, the bug starts to appear.

siduck:
> This bug happens on the latest commit of nvchad. The regression started at #818.

@zbirenbaum

siduck:
https://user-images.githubusercontent.com/59060246/154811373-ac969c65-5f6f-48ee-9f69-24616d1aedcb.mp4

works fine here :/

zbirenbaum:
[...]

Issue #824, opened 2022-02-14
Infinite loop caused by requiring plugin table


zbirenbaum:
While trying to debug what on earth was going on with my config for #823, I realized trying to load chadrc from my custom/plugins_table file to get the plugin_status variable causes an infinite loop, since chadrc requires plugins_table for setting the install variable. I fixed it by making plugins.status a separate file required by both of them rather than requiring each other, but this may be something that should be disclaimed in the docs.

Also, since the user plugin table can be a pretty large file, requiring it every time the config is loaded is quite inefficient. I was thinking it maybe better to make plugins.install a string, and just require it once inside utils or init.lua (which would also fix the above issue), but it would be a breaking change...

zbirenbaum:
> everything's fine here :l

Make a plugin file and put ‘require(“core.utils”).load_config().plugins.status’, (or whatever it is, I’m on mobile) add a plugin, and watch it mess up on packer sync…

siduck:
adding a plugin works for me, btw we already defined the plugin status for default plugins. Why would you want to define it for custom plugins?

zbirenbaum:
> adding a plugin works for me, btw we already defined the plugin status for default plugins. Why would you want to define it for custom plugins?

I have a complex setup. For instance, if I have cmp enabled, I know coq needs to be disabled, etc. My custom plugins are enabled and disabled frequently so having a unified table is helpful.

Also, for people like me with a 200 line plugin table, requiring it every time the config is loaded cannot be efficient... I just PR'd an opt in fix that doesn't break anything.

PR #833, opened 2022-02-20
fix autocmd breaking term plugins

zbirenbaum:
While taking a look at #829 and #824 I realized my local fix for toggleterm would show up as a change and it was kinda annoying. This is a quick fix allowing the user to disable the autocmd which makes nvchad not place nice with term plugins.

Really only temporarily needed until I have time to fully implement #811


siduck (approved):

siduck:
line numbers in terminal looks good, it must be enabled by default tho

siduck (approved):

zbirenbaum:
> line numbers in terminal looks good, it must be enabled by default tho

the autocmd does run by default.

siduck:
oh oops

PR #834, opened 2022-02-20
Plugin status fix

### LINKED ISSUE #824: Infinite loop caused by requiring plugin table

zbirenbaum:
Fixes #824 

This allows 'install' in chadrc to be a string, meaning that the plugin table won't be loaded every time the config is. This should give a light performance boost as well as fixing the breakage caused by requiring the config from the plugin table.

To prevent this from being a breaking change, table values and require("some.user.table") are still accepted. Users who wish to reference plugins.status from their config table, or boost their config's efficiency should use install = "some.user.table" instead though.

### DIFF
--- a/lua/core/autocmds.lua
+++ b/lua/core/autocmds.lua
@@ -1,3 +1,4 @@
+local settings=require("core.utils").load_config().options.nvChad
 -- uncomment this if you want to open nvim with a dir
 -- vim.cmd [[ autocmd BufEnter * if &buftype != "terminal" | lcd %:p:h | endif ]]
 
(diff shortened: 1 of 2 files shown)


zbirenbaum:
Oops the reversion to your commits earlier messed something up here. One sec I'll fix it.

zbirenbaum:
#835

PR #842, opened 2022-02-25
Make lsp keybindings take effect

lijunzh:
With the current config, the keybinding defined in `core/mapping` for lspconfig does not take effect. I don't see the clear benefit of defining these keybindings with the `on_attach` function. Thus, I propose to move it out to the same level as `lsp_handlers`.

### DIFF
--- a/lua/plugins/configs/lspconfig.lua
+++ b/lua/plugins/configs/lspconfig.lua
@@ -1,4 +1,5 @@
 require("plugins.configs.others").lsp_handlers()
+require("core.mappings").lspconfig()
 
 local function on_attach(_, bufnr)
    local function buf_set_option(...)
@@ -7,8 +8,6 @@ local function on_attach(_, bufnr)
 
    -- Enable completion triggered by <c-x><c-o>
    buf_set_option("omnifunc", "v:lua.vim.lsp.omnifunc")
-
-   require("core.mappings").lspconfig()
 end
 
 local capabilities = vim.lsp.protocol.make_client_capabilities()


lijunzh:
I am using [nvim-lsp-installer](https://github.com/williamboman/nvim-lsp-installer) to automatically manage the LSP server installation and configuration. Once switched to the manual configuration using `M.plugins.options.lspconfig.setup_lspconfig` defined in `custom/chadrc.lua`, the keybinding works. 

Maybe it still make sense to move it outside the `on_attach` function so that it will be compatible with other LSP managers. Otherwise, I am closing this PR as it is not necessary.

siduck:
> 

lspinstalle stuff will also work, you just need to add the on_attach in its config

siduck:
> @siduck Thanks a lot. I just realize that you can override the LSP server config for nvim-lsp-install in their [wiki page](https://github.com/williamboman/nvim-lsp-installer/wiki/Advanced-Configuration). Just put the link here for reference to others.

alright, anyways we have that mentioned as an example in our lsp install docs too!

image

lijunzh:
@siduck Thanks for the reference. I did look at your example before. However, I feel like some keybindings for basic operations, e.g. rename, formating, go to definition, and go to reference, are pretty universal for all LSP servers. It does not make sense to me to set them up manually for each LSP server like what is shown in the example. 

On the other hand, I did follow your example [here](https://github.com/siduck/dotfiles/blob/049376f767f3519b3fc7f4e6099dcf7970ef8b40/nvchad/custom/plugins/lspconfig.lua#L13-L24) to use the default `on_attach` function and setup my LSP. The downside is that the `pyright` LSP does not support `formatting` action, which I have to use [neoformat](https://github.com/sbdchd/neoformat) to call `black` to format my code. I can no longer use my beloved `<leader>fm` as it is taken by bindings set in LSP `on_attach`. Other than that, I am pretty happy with the current setup. 

Thanks again for your help.

siduck:
> @siduck Thanks for the reference. I did look at your example before. However, I feel like some keybindings for basic operations, e.g. rename, formating, go to definition, and go to reference, are pretty universal for all LSP servers. It does not make sense to me to set them up manually for each LSP server like what is shown in the example.
> 
> On the other hand, I did follow your example [here](https://github.com/siduck/dotfiles/blob/049376f767f3519b3fc7f4e6099dcf7970ef8b40/nvchad/custom/plugins/lspconfig.lua#L13-L24) to use the default `on_attach` function and setup my LSP. The downside is that the `pyright` LSP does not support `formatting` action, which I have to use [neoformat](https://github.com/sbdchd/neoformat) to call `black` to format my code. I can no longer use my beloved `<leader>fm` as it is taken by bindings set in LSP `on_attach`. Other than that, I am pretty happy with the current setup.
> 
> Thanks again for your help.

I use null-ls , you can check https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md#formatting 
https://www.reddit.com/r/neovim/comments/l3xm4f/how_to_format_python_code_when_using_pyright/
https://www.reddit.com/r/neovim/comments/kpkc7o/how_to_leverage_neovims_vimlspbufformatting/

lijunzh:
> I use null-ls , you can check https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md#formatting
> https://www.reddit.com/r/neovim/comments/l3xm4f/how_to_format_python_code_when_using_pyright/
> https://www.reddit.com/r/neovim/comments/kpkc7o/how_to_leverage_neovims_vimlspbufformatting/

@siduck This is very helpful. I've adopt the null-ls solution. It works better than `neoformat` and I can now chain both the `black` and `sort` together in the formatting command.

lijunzh:
I ended up defining the formatting keybinding in the `null-ls.lua`. Below is my version:

```lua
local null_ls = require("null-ls")
local b = null_ls.builtins

local sources = {

	-- lua
	b.formatting.stylua,
	b.diagnostics.luacheck.with({ extra_args = { "--global vim" } }),

	-- python
	b.formatting.black,
	b.formatting.isort,

	-- shell
	b.formatting.shfmt,
}

local M = {}

M.setup = function()
	null_ls.setup({
		debug = true,
		sources = sources,
		on_attach = function(client)
			if client.resolved_capabilities.document_formatting then
				vim.cmd([[
					nnmap <leader>fm :lua vim.lsp.buf.formatting_sync()<CR>
				]])
			end
		end,
	})
end

return M
```
So far, I have not yet had any comflicts with `lspconfig`'s `on_attach` function. But, I am not sure if this will works with other LSPs with formatting capabilities.

siduck:
@lijunzh i ave tried adding the formatting keybind in null-ls setup like yours but i get this prompt whenever formatting 
image
image

siduck:
this works https://github.com/siduck/dotfiles/blob/master/nvchad/custom/plugins/lspconfig.lua#L6

lijunzh:
> this works https://github.com/siduck/dotfiles/blob/master/nvchad/custom/plugins/lspconfig.lua#L6

@siduck Interesting. I tried your code first, but my other keybindings, e.g. go to definition, etc, stopped working. That's why I started to mess with `null-ls` and ended up with that solution. For the annoying message you show here


> @lijunzh i ave tried adding the formatting keybind in null-ls setup like yours but i get this prompt whenever formatting
> image
> image


I think you can get rid of it by the method described by `null-ls` [wiki page](https://github.com/jose-elias-alvarez/null-ls.nvim/wiki/Avoiding-LSP-formatting-conflicts). You do need to disable the `tsserver` formatting capability first, and probably get rid of the `if` statement on line 26 as it will always be `false` for `server`.

PR #862, opened 2022-03-06
More intuitive method for setting Neovim/Vim options. Less work for the user and developer.

zahimeen:
**The Following Is Simply An Advertisement For A Trash Product**

Do you want less work for your developers and users?
Well, switch to `for` loops now which are much more intuitive and requires much less work. Now your users no longer need to go through the hassle of making their own damn `lua` file just to load some options. And you no longer have to type out the damn option every single time yourself.

**Disclaimer**

Your customers will be angry at you because this breaks their config and they may need to rework a few things. For example, take 2 seconds to copy and paste their previous options into `M.options.nvim` in the `chadrc.lua` file.

As far as I can see, I have not noticed a startuptime difference (`nvim --startuptime random_file_i_forgot_about`).

By the way, this solves nothing at all!

**Another Disclaimer**

The mapleader option can now be found in `M.mappings.mapleader` in the `chadrc.lua` file for the user and `default_config.lua` file for the developers looking for it.

***Please don't ban me for opening an absolutely useless pull request. I swear I can contribute more than 2 lines of code.***

### DIFF
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -4,29 +4,34 @@
 local M = {}
 
 M.options = {
-   -- custom = {}
-   -- general nvim/vim options , check :h optionname to know more about an option
-
-   clipboard = "unnamedplus",
-   cmdheight = 1,
-   ruler = false,
-   hidden = true,
-   ignorecase = true,
-   smartcase = true,
-   mapleader = " ",
-   mouse = "a",
-   number = true,
-   numberwidth = 2,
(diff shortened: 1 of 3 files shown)


peppapig450:
? for loops will 100% increase startup time because loops take time to run

peppapig450:
This also completely goes against what the project has been about for a long time

siduck:
I agree with pepe

PR #865, opened 2022-03-06
fix(color): don't clear `CursorLine` highlight

### LINKED ISSUE #863: I can't enable cursor line
I added this to `custom/init.lua` to enable the cursor line but it didn't work:
```lua
vim.opt.cursorline = true
```
Then I noticed that the cursor line is enabled by default but for some reason, you cleared the `CursorLine` highlight in `colors/highlight.lua` so it doesn't show up.

```lua
-- Disable cursor line
cmd "hi clear CursorLine"
```

amirhhashemi:
fixes #863

### DIFF
--- a/lua/colors/highlights.lua
+++ b/lua/colors/highlights.lua
@@ -37,8 +37,6 @@ else
    fg("Comment", grey_fg)
 end
 
--- Disable cursor line
-cmd "hi clear CursorLine"
 -- Line number
 fg("cursorlinenr", white)
 


siduck:
thats not needed . I cleared the cursorline as it looks ugly but some need it so just override highlight

image

see this works. But you dont need to add highlight there. 

https://nvchad.github.io/config/Custom%20config#override-default-highlights

jeffvandyke:
In my opinion, this pull request should be re-opened and merged. There's no clear justification for such an unintuitive sweeping change like clearing the CursorLine highlight group, and it renders the `'cursorline'` option useless when enabled for visual purposes, leading to confusion and wasted time.

jeffvandyke:
I also added a note - https://github.com/NvChad/NvChad/pull/865#issuecomment-1785811684 - If some people don't like the CursorLine appearance, then the correct solution is: _disable the `'cursorline'` option_ - that's it's entire point is to show this line, and needing to do such convoluted workarounds to turn it _back_ on for the people that find it useful is time wasting.

Shahabaz-Bagwan:
> In my opinion, this pull request should be re-opened and merged. There's no clear justification for such an unintuitive sweeping change like clearing the CursorLine highlight group, and it renders the `'cursorline'` option useless when enabled for visual purposes, leading to confusion and wasted time.

Reason to clear highlight group, 
https://stackoverflow.com/questions/8247243/highlighting-the-current-line-number-in-vim

jeffvandyke:
That makes more sense now. Perhaps this PR should be amended by simultaneously setting the default option `cursorlineopt=number` to only highlight the number and not the line. It seems to be a recent option, and I can't find it previously mentioned in any PRs or issues in this repo.

siduck:
@jeffvandyke added in v3.0!

PR #889, opened 2022-03-20
Header has too much padding.

Johnsiras:


siduck:
that depends on the user's screen size i think 

padding 4 

image

padding 9
image

siduck:
so we need to make the padding dynamic ._.

siduck:
how does this look

```
layout = {
         { type = "padding", val = 9 },
         default.header,
         { type = "padding", val = 2 },
         default.buttons,
         { type = "padding", val = 9 },
      },
      ```

Johnsiras:
can't we make like check the user's screen size then put it there in padding?

Johnsiras:
> how does this look
> 
> ```
> layout = {
>          { type = "padding", val = 9 },
>          default.header,
>          { type = "padding", val = 2 },
>          default.buttons,
>          { type = "padding", val = 9 },
>    },
> ```

wezterm-gui_woWfFoZwKf

siduck:
> can't we make like check the user's screen size then put it there in padding?

https://github.com/goolord/alpha-nvim/issues/93#issuecomment-1073416872

PR #899, opened 2022-03-24
fix nvimtree vim globals not having any effect

zbirenbaum:
Currently, changing any of the options stored in vim globals has no effect. This can be easily reproduced by attempting to disable folder icons by changing the code in the default file like so:
```
g.nvim_tree_show_icons = {
   folders = 1,
   files = 1,
   git = 1,
}
```
to 
```
g.nvim_tree_show_icons = {
   folders = 0,
   files = 0,
   git = 1,
}
```
The icons will still be there on restart. With the change to move the setting globals above the require statement in this patch, the above code functions properly.

### DIFF
--- a/lua/plugins/configs/nvimtree.lua
+++ b/lua/plugins/configs/nvimtree.lua
@@ -1,9 +1,4 @@
-local present, nvimtree = pcall(require, "nvim-tree")
-
-if not present then
-   return
-end
-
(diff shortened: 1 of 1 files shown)


Johnsiras:
Does this works even without setting the `lazy_load` nvimtree option in chadrc to false?

Johnsiras:
And yea bout it, does this fix the issue 🤔

Issue #631, opened 2021-12-07
custom theme dir


siduck:
It would be great if we added a custom theme dir path as an option in the chadrc 

something like this : 

![image](https://user-images.githubusercontent.com/59060246/139169176-bf818399-cdab-4798-9805-a271ceda18b6.png)

And the custom folder must have the user's theme folders , which must have 2 folders : 

hl_themes 
themes 

![image](https://user-images.githubusercontent.com/59060246/139169109-3db3554b-9319-49c4-8590-b6d3e64ddf99.png)

- changes must be done in the telescope theme switcher

siduck:
Ive noticed this a lot that I have to intervene and edit themes personally when theme PR's come which makes me think that y'all wouldnt be able to make perfect themes for nvchad! so I dont think this 'custom' local theme feature would be of any use. just send new theme requests , I will make them when free.

tahsin-npx:
Is it possible to use .vim theme files with Nvchad? :/

siduck:
> Is it possible to use .vim theme files with Nvchad? :/

ofc, just source it. 

```
vim.cmd " ~/.config/nvim/lua/custom/siduck.vim"
```

tuberry:
It seems that there only are two ways to use and modify private imperfect themes:
 1) submit the theme to the repository and wait for review,
 2) `so` the theme in the `custom` folder and undo all the redundant parts of the default theme.

IMHO, they aren't that "customizable".

siduck:
1

> It seems that there only are two ways to use and modify private imperfect themes:
> 
> 1. submit the theme to the repository and wait for review,
> 2. `so` the theme in the `custom` folder and undo all the redundant parts of the default theme.
> 
> IMHO, they aren't that "customizable".

1 option is only viable

tuberry:
> 1 option is only viable

They look more like multiple color schemes within the same [theme](https://github.com/NvChad/nvim-base16.lua/blob/master/lua/base16.lua) rather than multiple themes. I would prefer a custom theme rather than customizing a theme.

siduck:
> > 1 option is only viable
> 
> They look more like multiple color schemes within the same [theme](https://github.com/NvChad/nvim-base16.lua/blob/master/lua/base16.lua) rather than multiple themes. I would prefer a custom theme rather than customizing a theme.

wym same theme?

tuberry:
Why not let users set any theme in a more convenient way? i.e.
```diff
[...]

PR #907, opened 2022-03-25
feat: support themes other than defaults

tuberry:
With this pr, users could use any themes they want. Specifically, 
1.  add themes to `userPlugins`
```lua
--  custom/plugins/inited.lua
...
{ "local/path/to/mytheme" }, -- local
{ "Iron-E/nvim-highlite" }, -- remote
...
```
2.  add colors file (i.e. `custom/colors.lua`) in format of [this](https://github.com/NvChad/nvim-base16.lua/blob/master/lua/hl_themes/aquarium.lua)
3.  set colors and theme in `chadrc.lua`
```lua
-- custom/chadrc.lua
...
M.ui = {
  theme = "mytheme",
  -- theme = "highlite",
  colors = "custom.colors",
}
...
```

Related to #631.

### DIFF
--- a/lua/colors/init.lua
+++ b/lua/colors/init.lua
@@ -13,22 +13,34 @@ M.init = function(theme)
 
    if present then
       -- first load the base16 theme
-      base16(base16.themes(theme), true)
+      local ok, array = pcall(base16.themes, theme)
 
-      -- unload to force reload
-      package.loaded["colors.highlights" or false] = nil
-      -- then load the highlights
-      require "colors.highlights"
+      if ok then
+         base16(array, true)
+         -- unload to force reload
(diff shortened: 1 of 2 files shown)


siduck:
the remote theme wouldnt be compatible with nvchad and it'd mess up some highlights

tuberry:
The purpose of this pr is not to be compatible with any theme but to provide a possibility to use individual themes, in this case theme compatibility should be the responsibility of the users themselves.

siduck:
@tuberry if you dont mind you could add lil docs for this feature over here https://github.com/NvChad/nvchad.github.io/blob/src/docs/config/Custom%20config.md , would ease my work!

Issue #925, opened 2022-04-01
Issues with new implementation of terminal toggling 


siduck:
@zbirenbaum 

## 1 

- Terminals opened on right/bottom i.e like split windows are listed in the bufferline 

![image](https://user-images.githubusercontent.com/59060246/161195687-c0036815-1187-4e0f-a747-eb49aadb74c2.png)

Old behaviour : they werent listed in the bufferline , nobufflisted stuff

zbirenbaum:
> @zbirenbaum 
> 
> 
> 
> 1. terminals opened on right/bottom i.e like split windows are listed in the bufferline 
> 
> 
> 
> ![image](https://user-images.githubusercontent.com/59060246/161195687-c0036815-1187-4e0f-a747-eb49aadb74c2.png)
> 
> 
> 
> Old behaviour : they werent listed in the bufferline , nobufflisted stuff 

Ahh I think that's related to my fix for the telescope thing, didn't notice cuz I don't use buffer line. I'll take a look at it, thanks for letting me know.

siduck:
# 2

We can no longer spawn multiple terminals :c  

## Steps to reproduce 

- Press leader+v 2 times
- Old behaviour would spawn 2 vertical terminals but now it just minimizes the terminal buffer i.e pressing leader +v again will not show the terminal buffer but it will show it in the bufferline 

https://user-images.githubusercontent.com/59060246/161196111-39216c60-31d3-436e-9e9d-d6fd656647f8.mp4

Expected behaviour : leader + v / x should spawn multiple terminals rather than toggling

@zbirenbaum

siduck:
# 3 

Horizontal terminal eats up the area of nvimtree too, this shouldnt be happening 

![image](https://user-images.githubusercontent.com/59060246/161196523-baeb00e4-30ed-4743-b85e-f06bb9a73cb8.png)

zbirenbaum:
> # 2
> 
> 
> 
> We can no longer spawn multiple terminals :c  
> 
> 
> 
> ## Steps to reproduce 
> 
> 
> 
> - Press leader+v 2 times
> 
> - Old behaviour would spawn 2 vertical terminals but now it just minimizes the terminal buffer i.e pressing leader +v again will not show the terminal buffer but it will show it in the bufferline 
> 
> 
> 
> https://user-images.githubusercontent.com/59060246/161196111-39216c60-31d3-436e-9e9d-d6fd656647f8.mp4
> 
> 
> 
[...]

PR #927, opened 2022-04-02
add new location option for chadterm to default config

zbirenbaum:
Chadterm will no longer overlap with nvimtree by default, but people who want to use the old splitting method or others can do so by setting the horizontal and vertical fields in options.terminal.location from chadrc.

NvChad/extensions#11 has been PR'd and fixes the issues with bufferline and nvimtree mentioned in #925

### DIFF
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -48,6 +48,10 @@ M.options = {
          vsplit_ratio = 0.5,
          split_ratio = 0.4,
       },
+      location = {
+         horizontal = "rightbelow",
+         vertical = "rightbelow",
+      },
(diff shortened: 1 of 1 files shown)


siduck:
@zbirenbaum I had invited you to the org so you could make such fix changes quickly without waiting for me to review for it

zbirenbaum:
> @zbirenbaum I had invited you to the org so you could make such fix changes quickly without waiting for me to review for it 

Thanks for the invite, happy to be a part of it! Small things like this I'll go ahead and merge in then.

I still think having a second look for larger changes is a good idea if that's ok with you

PR #933, opened 2022-04-07
fix: remove invalid packer.nvim option

ahhshm:
Is `wants` a valid option? I can't find any documentation about it.

### DIFF
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -19,9 +19,9 @@ local plugins = {
 
    {
       "NvChad/extensions",
-      config = function ()
+      config = function()
          vim.schedule_wrap(require("nvchad.terminal").init())
-      end
+      end,
    },
 
    {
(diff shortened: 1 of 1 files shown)


bryant-the-coder:
change to `requires = "rafamadriz/friendly-snippets"`

siduck:
> Is `wants` a valid option? I can't find any documentation about it.

i got it from folke's config before

ahhshm:
yeah, I've seen it in some configurations but there is no documentation for that. It's probably deprecated. Give me some time to make sure about that.

ahhshm:
@siduck I found some clues about why it's not documented. 
Apparently, it was added to temporarily fix an issue with lazy loading and it wasn't supposed to remain that long. and I think it's not documented to avoid people using it.

This is the PR that added it: https://github.com/wbthomason/packer.nvim/pull/264
And this is the request to add documentation for it (for more information): https://github.com/wbthomason/packer.nvim/issues/537

I'm still not sure what exactly it does. I would suggest closing this PR and keeping the option. But you might want to merge it if you don't want to mess with its deprecation or its changes in behavior which is very likely.

PR #934, opened 2022-04-07
refactor: bad module name for `friendly-snippets`

ahhshm:
I don't know the reason behind that naming but It's a bad name. I think "friendly-snippets" is good enough.


siduck:
@ahhshm thats required! in case someone adds this in their lspconfig file
image , cmp would automatically load itself

PR #937, opened 2022-04-09
add disable option for statusline hide autocmd

zbirenbaum:
Description in title. Discovered when trying to test out global statusline that statusline_hide() always sets laststatus regardless of user options, making global statusline impossible to enable. 

I I thought about adding a specific global option, but that's up to you @siduck, since I know you are trying to keep down bloat. Regardless, there should never be a case where we have an autocmd set an option and not have an option to disable it.

### DIFF
--- a/lua/core/autocmds.lua
+++ b/lua/core/autocmds.lua
@@ -12,7 +12,9 @@ if not settings.terminal_numbers then
 end
 
 -- Don't show status line on certain windows
-vim.cmd [[ autocmd BufEnter,BufRead,BufWinEnter,FileType,WinEnter * lua require("core.utils").hide_statusline() ]]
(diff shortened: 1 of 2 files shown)


siduck:
@zbirenbaum the global statusline changes will be done after 0.7 is released and by what I can see here https://github.com/NvChad/NvChad/issues/922, it seems like almost 90% of the users do not want the custom dir so the option provided in this PR will not be used ig

zbirenbaum:
@siduck We still need an option to disable that autocmd so I don't have to deal with that change every time I want to make a PR.

zbirenbaum:
Ty :) look forward to floating terminals hopefully ready in an hour or two.

PR #942, opened 2022-04-10
Fix the insert mode nav issue

ccwanggl:
I think the `next_line` and `prev_line` are configured to the contrary.


siduck:
@ccwanggl use these lines in mappings.lua too
```
   map("i", inav.next_line, "<Down>")
   map("i", inav.prev_line, "<Up>")
```

PR #967, opened 2022-04-24
Mappings rewrite

zbirenbaum:
Pr so it's easier to review.

@Akianonymus commented on the recursion being unnecessary, but I found it necessary to include otherwise it received an error. Let me know if that's just me and I'll commit amend it though.

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -127,34 +127,18 @@ M.misc = function()
       -- terminal mappings end --
 
       -- Add Packer commands because we are not loading it at startup
-      user_cmd("PackerClean", function()
-         require "plugins"
-         require("packer").clean()
-      end, {})
-      user_cmd("PackerClean", function()
-         require "plugins"
-         require("packer").clean()
-      end, {})
-      user_cmd("PackerCompile", function()
-         require "plugins"
-         require("packer").compile()
-      end, {})
-      user_cmd("PackerInstall", function()
-         require "plugins"
-         require("packer").install()
-      end, {})
-      user_cmd("PackerStatus", function()
-         require "plugins"
-         require("packer").status()
-      end, {})
-      user_cmd("PackerSync", function()
-         require "plugins"
-         require("packer").sync()
-      end, {})
-      user_cmd("PackerUpdate", function()
-         require "plugins"
-         require("packer").update()
-      end, {})
+      local packer_cmd = function(callback)
+        return function()
+          require "plugins"
+          require("packer")[callback]()
+        end
+      end
+      user_cmd("PackerClean", packer_cmd("clean"), {})
+      user_cmd("PackerCompile", packer_cmd("compile"), {})
+      user_cmd("PackerInstall", packer_cmd("install"), {})
+      user_cmd("PackerStatus", packer_cmd("status"), {})
+      user_cmd("PackerSync", packer_cmd("sync"), {})
+      user_cmd("PackerUpdate", packer_cmd("update"), {})
 
       -- add NvChadUpdate command and mapping
       cmd "silent! command! NvChadUpdate lua require('nvchad').update_nvchad()"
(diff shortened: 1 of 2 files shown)


zbirenbaum:
> how about making a wrapper above with `local map_command = vim.api.nvim_create_user_command`, or any other name, just to reduce the code.

Done.

Akianonymus:
> I did need the recursion. It was added to fix an error. Unless that was added since my last recompile which was ~1 week ago.

I wil look into it.

Akianonymus:
Hmm, seems like upstream only supports multiple map modes, doesn't support multiple lhs.

siduck:
@zbirenbaum ready to merge?

Akianonymus:
Not working as intended, @zbirenbaum what's your neovim version?

siduck:
> Not working as intended, @zbirenbaum what's your neovim version?

weird, works here fine even on my other branch (re-custom)

zbirenbaum:
@Akianonymus I build from source every few days or so.
@siduck It has seemed to work fine for me too.

siduck:
https://github.com/MordechaiHadad/bob

Akianonymus:
For example if press escape key, which is mapped to :noh , :noh appears below the status line. Same for any other command.

My neovim info:
```
❯ vi --version
NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

zbirenbaum:
I just pushed and update which fixes the echo issue with comments, and dramatically cleans up the lsp and packer section. I'll  take a look to see if noh is doing the same and look over anything else that may be.

siduck:
image

yea i notice the same with nvimtreetoggle command :thinking:

zbirenbaum:
@siduck @Akianonymus It was really pervasive, I think I messed up a bulk replace somewhere. I fixed it with a regex search and replace so everything should be fine now.

lua/core/mappings.lua
```diff
@@ -152,56 +155,59 @@ end
...
-   map("n", m.prev_buffer, ":BufferLineCyclePrev <CR>")
+   map("n", m.next_buffer, "<cmd>:BufferLineCycleNext <CR>")
+   map("n", m.prev_buffer, "<cmd>:BufferLineCyclePrev <CR>")
 end
 
 M.comment = function()
    local m = plugin_maps.comment.toggle
-   map("n", m, ":lua require('Comment.api').toggle_current_linewise()<CR>")
-   map("v", m, ":lua require('Comment.api').toggle_linewise_op(vim.fn.visualmode())<CR>")
+   map("n", m, "<cmd>:lua require('Comment.api').toggle_current_linewise()<CR>")
```
Akianonymus, line 164: shouldn't this be moved to function syntax ? rather than :lua ....

zbirenbaum: I'm not sure how to get the range stuff working with lua commands. Putting it in the lua format broke the commenting. I can check if there's a easy way though.

zbirenbaum: There does not seem to be one. This plugin's source code is an absolute mess of vimscript. I could implement a better one in lua if one doesn't exist. For now, leave it as is. I'll commit with a fix if I find one. That, or I'll find a better plugin.

zbirenbaum: Wait i gotta commit amend this, group comment toggle is broken... again...

https://github.com/b3nj5m1n/kommentary/tree/main/lua looks like a good choice.

Akianonymus: Just remove the <cmd> from front, it will work but will just print the commands too, it wasn't working before or something ?

zbirenbaum: So the way I had it was exactly as it was before. Apparently it behaves differently with vim.keymap. This absolutely gross mapping is what's in the plugin's docs and that string finally seems to work so this is what we will do for now.

vim.keymap.set('x', '<C-_>', '<ESC><CMD>lua require("Comment.api").toggle_linewise_op(vim.fn.visualmode())<CR>')

Even the other plugin I linked seems to have similar issues. These people know that you can get marks from lua right? They don't have to get ranges through the vim api, but all of them do.... I checked 3 different plugins. Absolutely ridiculous. I may have to write the 10 millionth comment plugin to finally have a sane api.

siduck: @zbirenbaum switch to a better commenting plugin instead

zbirenbaum: @siduck see discussion above. I checked around. There are better(ish) ones but they actually don't seem to preserve our current commenting behavior. I think this should be merged and I can look into it later.

Judging by the api's the 3 I checked out offer, I may have to write it myself.

siduck: @zbirenbaum what about this? https://github.com/terrortylor/nvim-comment

siduck:
already included in re-custom implementation

PR #983, opened 2022-04-28
added gc and gb as keys to trigger lazy load on comment plugin

primeapple:
The `Comment.nvim` plugin now correctly lazy loads. :)
Before it did not lazy load for things like `gcaw`, `gbaw` in normal mode or `gc` in visual mode.

### DIFF
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -192,7 +192,7 @@ local plugins = {
       "numToStr/Comment.nvim",
       disable = not plugin_settings.status.comment,
       module = "Comment",
-      keys = { "gcc" },
+      keys = { "gc", "gb" },
       config = override_req("nvim_comment", "plugins.configs.others", "comment"),
       setup = function()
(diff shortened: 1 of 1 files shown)


siduck (approved):

siduck:
@primeapple send the PR again, idk why im not able to merge it. Where's the conflicts?

primeapple:
Will send it again later this day :)

peterpme:
:wave: Is the problem here the `gcc` commands vs. `gc`, `gb`? For example, I use `gcc` myself - is that the trade-off?

siduck:
cant we have gcc and gc both at the same time?

primeapple:
The thing is, that `gc` implies `gcc` ;-)  
So as soon as someone types `gc` the plugin is being loaded.

We have to get a list of all the used prefixes and put them into this list. For `Comment.nvim` this is `gc` and `gb`.

primeapple:
See #997

PR #992, opened 2022-04-29
set LSP mappings only for buffers that have a configured languag server

ahhshm:


siduck (approved):

lua/core/mappings.lua
```diff
@@ -96,73 +96,79 @@ M.comment = function()
...
 
-M.lspconfig = function()
+M.lspconfig = function(bufnr)
    -- See `<cmd> :help vim.lsp.*` for documentation on any of the below functions
-   map("n", "gD", function()
+
+   local buf_map = function(...)
+      local key, lhs, rhs = ...
+      map(key, lhs, rhs, { buffer = bufnr })
+   end
```
ahhshm, line 105: This could be in the utils file but I don't think it is used anywhere else + it needs some domifications

lua/core/mappings.lua
```diff
@@ -96,73 +96,79 @@ M.comment = function()
    map("v", "<leader>/", "<esc><cmd> :lua require('Comment.api').toggle_linewise_op(vim.fn.visualmode())<CR>")
 end
 
-M.lspconfig = function()
+M.lspconfig = function(bufnr)
    -- See `<cmd> :help vim.lsp.*` for documentation on any of the below functions
-   map("n", "gD", function()
+
+   local buf_map = function(...)
+      local key, lhs, rhs = ...
```
ahhshm, line 103: This probably can be simplified with Lua trick that I don't know

siduck: no worries, its alright for the time being

PR #997, opened 2022-04-30
New try of #983, as well removing unnecessary opt key

primeapple:
See #983

Also the `opt` key is implied by the `setup` key, so it is not needed.

### DIFF
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -78,6 +78,7 @@ local plugins = {
 
    -- git stuff
    ["lewis6991/gitsigns.nvim"] = {
+      opt = true,
       config = function()
          require("plugins.configs.others").gitsigns()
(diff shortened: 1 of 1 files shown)


siduck:
@primeapple opt is needed for lazy loading!

primeapple:
Try it out. It should be fine, because packer sets `opt` automatically when using `setup`. At least when I tried it last it worked fine! :)

siduck:
i tried it and having opt definitely makes some changes. 

check the startuptime

image

siduck:
ig our lazy load function needs the plugin to be in opt dir ._.

primeapple:
I'll put it back for now. However, I have an idea to remove the lazy load function alltogether. Packer should be able to do this with `event = 'VimEnter'`

PR #1022, opened 2022-05-04
clearer display of open folders in nvimtree

brutusmcforce:
Open folders are confusing since they indent to a level which visually indicates they are a subfolder of the folder above them in the tree.

For comparison, before this change:
before

After this change:
after


siduck:
what about this? 

image

brutusmcforce:
For visibility, even though we're discussing on Discord too:

I don't think there should ever be any case where the icons should be wider, or more icons should be added (and thus increasing width) in a tree structure. It very easily gets confusing, especially when navigating larger file structures.

I don't think the open folder icon is ugly!

PR #1041, opened 2022-05-08
Fix deprecated client_capabilities usage

DanisDGK:
`client_capabilities` is now deprecated, `server_capabilities` does the same thing.


DanisDGK:
Both work in 0.7 as far as I'm aware, by the way, however apologies if I am wrong about that

PR #1042, opened 2022-05-08
BREAKING CHANGE: Rewrote lsp server configuration & made it modular

LeonHeidelbach:
This is a breaking change but it is necessary to enable users to configure their LSPs in a modular way and it also makes the configuration part a lot easier. LSP servers that are installed using `LspInstall` or `LspInstallInfo` will now also be loaded with the default configuration even if they are not listed in the lspconfig's `M.servers` field. @siduck try it out and let me know what you think of this.

This is my current lspconfig.lua which stands as an example for how you would configure your LSPs.

```lua
-- $HOME/.config/nvim/lua/custom/plugins/lspconfig.lua

-- Here you can define your own setup parameters for each LSP as well as override the default configs

local LSP = {}

LSP.default_config = {
  on_attach = function(client, bufnr)
    -- enable document formatting
    client.server_capabilities.document_formatting = true
    client.server_capabilities.document_range_formatting = true

    -- if document formatting is enabled, set a keybinding to format the file or a selection
    if client.server_capabilities.document_formatting then
      vim.api.nvim_buf_set_keymap(bufnr, "n", "<space>fm", "<cmd>lua vim.lsp.buf.formatting()<CR>", {})
      vim.api.nvim_buf_set_keymap(bufnr, "v", "<space>fM", "<cmd>lua vim.lsp.buf.range_formatting()<CR>", {})

      -- format the file on save
      vim.cmd "autocmd BufWritePre <buffer> lua vim.lsp.buf.formatting_sync()"
    end

    -- print if the language server supports formatting
    print("LSP formatting enabled:", client.server_capabilities.document_formatting)

    -- require the default keybindings
    require("core.mappings").lspconfig()
  end,
}

-- This is only supposed to be an example on how to create a custom LSP configuration.
-- LSP.sumneko_lua = {
--   on_attach = function(client, _)
--     -- Put your code in here
--     LSP.default_config.on_attach(client, _)
--   end,
-- }

local M = {}

M.setup_lsp = function(lsp)
  -- here you can return the configurations for your LSP servers. The returned table will be
  -- deep-merged with the default one, which means that you can also override only parts of
  -- the default config (e.g. see above, LSP.sumneko_lua only changes the on_attach callback).
  -- IMPORTANT: if you would like to retain the default capabilities for a new custom server setup,
  -- you need to set the capabilities field.

  -- set the default capabilities to all custom LSP configs
  for k, _ in pairs(LSP) do
    LSP[k].capabilities = lsp.default_config.capabilities
  end

  return LSP
end

-- The list of LSP servers you would like to install and use by default.
-- Servers that are installed through :LspInstall will be automatically added to this list
-- automatically.
M.servers = {
  "jdtls",
  "pylsp",
  "clangd",
  "cssls",
  "html",
  "jsonls",
  "texlab",
  "sumneko_lua",
  "tsserver"
}

return M
```

### DIFF
--- a/lua/plugins/configs/lspconfig.lua
+++ b/lua/plugins/configs/lspconfig.lua
@@ -1,19 +1,11 @@
-local present, lspconfig = pcall(require, "lspconfig")
+local lsp_config_present, lspconfig = pcall(require, "lspconfig")
+local lsp_installer_present, lsp_installer = pcall(require, "nvim-lsp-installer")
 
-if not present then
-   return
+if not lsp_config_present or not lsp_installer_present then
+  return
 end
-
-local M = {}
 
 require("plugins.configs.others").lsp_handlers()
-
-function M.on_attach(client, _)
-   client.resolved_capabilities.document_formatting = false
-   client.resolved_capabilities.document_range_formatting = false
-
-   require("core.mappings").lspconfig()
-end
 
 local capabilities = vim.lsp.protocol.make_client_capabilities()
 capabilities.textDocument.completion.completionItem.documentationFormat = { "markdown", "plaintext" }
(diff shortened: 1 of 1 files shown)


siduck:
@LeonHeidelbach ig this will add more complexity and it would be good to stay with the current one as many get annoyed due to breaking changes. Also server capabilities didnt work for me in 0.7 but resolved capabilities did. Closing this

LeonHeidelbach:
> @LeonHeidelbach ig this will add more complexity and it would be good to stay with the current one as many get annoyed due to breaking changes. Also server capabilities didnt work for me in 0.7 but resolved capabilities did. Closing this

@siduck No this actually removes a lot of complexity for the user and enables them to specifically override or add the things they want to change. It also allows for an easy way to configure on a per LSP basis as well as allows for configuring multiple LSPs the same way. I know users get annoyed by frequent breaking changes but with this they would only have to remove unnecessary code from before. This solution would also make it a lot easier for new users to configure LSPs. Also the old way did just simply make `LspInstall` and `LspInstallInfo` useless since the servers installed through these methods would just not be loaded unless they were manually added to the lspconfig.

siduck:
@LeonHeidelbach it will be good to adhere to the defaults as in configuration of lspinstaller as the installed servers are meant to be configured with lspconfig. The current setup is easy too btw ._.

LeonHeidelbach:
@siduck and regarding the "server_capabilities" I have left it at "resolved_capabilities" in my commit and just changed it in my own default_config override as you can see

LeonHeidelbach:
> 

They are still configured through the lspconfig.

LeonHeidelbach:
@siduck "nvim-lsp-installer" is only used to get the list of installed lsps. I could change it so that this all also works if "nvim-lsp-installer" is not installed that would only be a small change.

siduck:
ik, but still! this makes the config looks cluttered with all the comments and all

LeonHeidelbach:
> ik, but still! this makes the config looks cluttered with all the comments and all,current lspconfig looks much cleaner imo

@siduck The config I posted was supposed to be an example with the comments only explaining everything that can be done. The minimum config could be as much as this:

```lua
local M = {}

M.setup_lsp = function(lsp)
  for k, _ in pairs(LSP) do
    LSP[k].capabilities = lsp.default_config.capabilities
  end
  return LSP
end

M.servers = {
  "jdtls",
  "pylsp",
  "clangd",
  "cssls",
  "html",
  "jsonls",
  "texlab",
  "sumneko_lua",
  "tsserver"
}

return M
```

And that's it :smile:. I think that's a lot cleaner than the current way users have to setup their lspconfig.lua.

LeonHeidelbach:
@siduck I could even set it up so that users would not even have to define `M.setup_lsp`at all and just set their list of servers in `M.servers` as an absolute minimum setup.

siduck:
@LeonHeidelbach ik but this would add more bloat xD JK! let the user do some lil effort too. Lets no spoonfeed and spoil the babies!

LeonHeidelbach:
@siduck I mean it's not even about that. It's just that there is no modularity in setting up the lspconfig so that for example only a couple of things can be changed from the default config i.e you set up `sumneko_lua` straight away with you default config. Let's say I want to change the capabilites for that language only. As of now I cannot. This PR would allow to organize and customize parts of predefined configs and only override stuff if needed. That's for advanced users only though and babies could only use the list of servers, so it's the best of both worlds :wink:. But, oh well if you think this is not the way to go then that's that.

sherlock1453:
It would be really appreciated, if you can add the custom configuration required for LSP in the examples directory. Can I open a separate issue for this ?

siduck:
@sherlock1453  https://nvchad.github.io/config/Lsp%20stuff

siduck:
> It would be really appreciated, if you can add the custom configuration required for LSP in the examples directory. Can I open a separate issue for this ?

this change wasnt made so dont worry about it

PR #1054, opened 2022-05-10
refactor: add utils to the global scope

### LINKED ISSUE #1053: Why the utility functions are not in the global scope?
Why the utility functions are not in the global scope? 90% (if not all) of the files use at least one of the utility functions and having them in the global scope eliminates a lot of `require`s and makes the code base a lot cleaner.

ahhshm:
closes #1053

I will do the same thing to the extensions and the base46 plugins

### DIFF
--- a/examples/init.lua
+++ b/examples/init.lua
@@ -1,7 +1,7 @@
 -- example file i.e lua/custom/init.lua
 
 -- MAPPINGS
-local map = require("core.utils").map
+local map = nvchad.map
 
 map("n", "<leader>cc", ":Telescope <CR>")
 map("n", "<leader>q", ":q <CR>")
--- a/init.lua
+++ b/init.lua
@@ -5,6 +5,7 @@ if present then
 end
 
 local core_modules = {
+   "core.utils",
    "core.options",
    "core.autocmds",
    "core.mappings",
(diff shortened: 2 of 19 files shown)


siduck:
@ahhshm why nvchad.map? cant we use map? we can just load core.utils in related files

ahhshm:
Yes, it's probably cleaner but I think it should be clear where the `map` function comes from for example.

ahhshm:
> _G syntax looks weird to me

we can define a global variable without `_G` but it has some benefits:
> One advantage of this structure is that it simplifies the internal implementation of Lua, because there is no need for a different data structure for global variables. The other (actually the main) advantage is that we can manipulate this table as any other table. To facilitate such manipulations, Lua stores the environment itself in a global variable _G. (Yes, _G._G is equal to _G.)

siduck:
I just didnt know how the _G stuff worked so got confused

ahhshm:
> I just didnt know how the _G stuff worked so got confused

I don't know either. I just know how to use it :)

Akianonymus:
Please do revert this, having global functions have a lot of bad implications, should be avoided as much as possible.

siduck:
> Please do revert this, having global functions have a lot of bad implications, should be avoided as much as possible.

wym :thinking:

Akianonymus:
Well more memory usage, but to be honest, memory usage is not a realistic concern here i guess. It's much more important that the code be easy to follow and not have unpredictable results. Global functions are also globally modifiable too, so there's is that. But anyway, ignore me.

PR #1066, opened 2022-05-12
feat: added which-key keymaps, fixed telescope keymap & custom "mode_opt" for keymaps

LeonHeidelbach:
@siduck As the title suggests, I fixed some existing keymaps, added ones for which-key and added a feature where users can now set `mode_opts` per keymap. This is important for registering expressions or similar, however setting `mode_opts` is completely optional.

These are some of the `mode_opts` values that can be set:

```lua
-- IMPORTANT: `mode_opts` values will apply to all key mappings in `M.<YOUR_KEY_MAP_LIST>`

mode_opts =  {
  buffer = nil, -- Global mappings. Specify a buffer number for buffer local mappings
  silent = true, -- use `silent` when creating keymaps
  noremap = true, -- use `noremap` when creating keymaps
  nowait = false, -- use `nowait` when creating keymaps
  -- and more ...
}
```

Before it was not possible to add expressions or alter any other options on an extension basis. This can now be done like this

**NOTE**: `expr = true` will apply to all keymaps for `M.copilot`. If you only want some keymaps to have the specified `mode_opts` values, simply create another `M.<YOUR_KEY_MAP_LIST>` for the same extension (e.g. `M.copilot_2`):

```lua
-- Github Copilot Bindings
-----------------------------------------------------------
M.copilot = {
  mode_opts = { expr = true },
  i = {
    ["<C-H>"] = { 'copilot#Accept("<CR>")', "   copilot accept" },
  },
}
```

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -224,11 +224,11 @@ M.telescope = {
       -- find
       ["<leader>ff"] = { "<cmd> Telescope find_files <CR>", "  find files" },
       ["<leader>fa"] = { "<cmd> Telescope find_files follow=true no_ignore=true hidden=true <CR>", "  find all" },
-      ["<leader>fw"] = { "<cmd> Telescope live_grep <CR>", "  live grep" },
+      ["<leader>fw"] = { "<cmd> Telescope live_grep <CR>", "   live grep" },
       ["<leader>fb"] = { "<cmd> Telescope buffers <CR>", "  find buffers" },
       ["<leader>fh"] = { "<cmd> Telescope help_tags <CR>", "  help page" },
-      ["<leader>fo"] = { "<cmd> Telescope oldfiles <CR>", "  find oldfiles" },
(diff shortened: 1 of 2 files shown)


siduck (approved):

LeonHeidelbach:
@siduck I think we should add what I mentioned in my comment on how this all works to the website's documentation.

siduck:
@LeonHeidelbach add a section after this https://github.com/NvChad/nvchad.github.io/blob/src/docs/config/Custom%20config.md#override-default-mappings

LeonHeidelbach:
@siduck you want me to do that or do you want to add that yourself? I could do later today, so it will be a little until the.

siduck:
> @siduck you want me to do it or do you want to add that yourself? I could do it later today, so it will be a little until then.

yep I want you to do it :p

LeonHeidelbach:
Okay will do 😂👍🏻

Issue #1071, opened 2022-05-13
some of mappings override/disable not working


GreatestOfAllTimee:
### **Describe the bug**
we already discuss this on Discord. I disable whichkey and add `nvchad.no_WhichKey_map()` in `custom/init.lua`.

**The mapping below does not work**. I believe there is more mappings not able to configure:

```lua
local M = {}

...

M.nvterm = {
   n = {
      -- disable both mappings
      ["<A-h>"] = "",
      ["<A-v>"] = "",
   },
}

M.move = {
   v = {
      ["J"] = { "<cmd>m '>+1<CR>gv=gv" },
      ["K"] = { "<cmd>m '<-2<CR>gv=gv" },
   },

   n = {
      -- doesn't work even tho we already disable on M.nvterm
      ["<A-h>"] = { "<cmd> vertical resize +2 <CR>" }, 
      ["<A-l>"] = { "<cmd> vertical resize -2 <CR>" },
   }
}

...

return M
```

### **Expected behavior**
Should work as we expected.

### **Desktop (please complete the following information):**
 - Ubuntu 20.04 LTS
 - Wezterm & Kitty
 - NVIM v0.7.1-dev

### **Additional context**
I used Nvim v0.7.1-dev before latest breaking changes (https://github.com/NvChad/NvChad/commit/932bd6a74c968df5152ab53ad4ec37ec23339044) and it work fine.

GreatestOfAllTimee:
> @GreatestOfAllTimee nvterm mappings can be disabled now, do a packersync

I will try later. Did you try apply others mappings? Its work?

siduck:
> > @GreatestOfAllTimee nvterm mappings can be disabled now, do a packersync
> 
> I will try later. Did you try apply others mappings? Its work?

i will test the others later but the nvterm mappings issue not disabling will work now

LeonHeidelbach:
@siduck another thing that just came to mind that I also noticed while rewriting my mappings yesterday is that it now also matters in which order the mapping configurations in `M.<YOUR_KEY_MAP>`s are used in the call  to `wk.register`. If you want to use a mapping for an extension that has already been claimed by another one that previously claimed it, overriding it will only work if the key map is loaded last (Or you would have to specifically add that mapping for a different extension to the one that claims it last, which would be kinda unintuitive if you know what I mean). We could circumvent this issue by looping through all mappings first, checking for a match with any custom override and remove those doubling entries from the table.

siduck:
[...]

PR #1072, opened 2022-05-13
fix: default mappings can now be properly disabled, independent of load order

LeonHeidelbach:
@siduck The title says it all. This implements what I outlined here https://github.com/NvChad/NvChad/issues/1071#issuecomment-1125925059.

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -31,6 +31,7 @@ nvchad.load_config = function()
       -- merge user config if it exists and is a table; otherwise display an error
       local user_config = require "custom.chadrc"
       if type(user_config) == "table" then
+         conf.mappings = nvchad.prune_key_map(conf.mappings, user_config.mappings, { "mode_opts" })
          conf = vim.tbl_deep_extend("force", conf, user_config)
       else
          error "User config (chadrc.lua) *must* return a table!"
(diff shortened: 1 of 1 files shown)


siduck:
dang soo many looop, I wanted to keep the utils.lua smol and simple :c

LeonHeidelbach:
> dang soo many loops! I wanted to keep the utils.lua smol and simple :c

@siduck Well unfortunately there is no other simpler way to check if there are doubling key bindings.

siduck:
> > dang soo many loops! I wanted to keep the utils.lua smol and simple :c
> 
> @siduck Well unfortunately there is no other simpler way to check if there are doubling key bindings.

we dont have to, the users must disable the keys in their original tables

LeonHeidelbach:
> > > dang soo many loops! I wanted to keep the utils.lua smol and simple :c
> > 
> > 
> > @siduck Well unfortunately there is no other simpler way to check if there are doubling key bindings.
> 
> we dont have to, the users must disable the keys in their original tables

@siduck that's the thing, it does unfortunately not just work like that. Currently it is dependent on the load order in which the key bindings are mapped. If the specific key binding you want to disable for one plugin, but want to use for another it with another plugin, it will be assigned the value that is last assigned to it in alphabetical order of the `M.<PLUGIN_NAME>` in your table. That means that the load order can just change when you add a new plugin to your mapping list. My prune function will resolve this issue. There is unfortunately no other way. Try it out yourself:

Go to your custom mappings list and add `M.a` and try to override a key for `M.nvimtree` or any other plugin that starts with a letter after "a". This will not work. Now change `M.a` to `M.z` and it should work. This is not expected behavior so it is necessary to completely remove overridden keys from the default table at the time when `wk.register` is called.

siduck:
@LeonHeidelbach  works here

image

LeonHeidelbach:
@siduck try moving it before `M.nvimtree` maybe it's not in alphabetical order but just assignment order. I had this issue multiple times and tried a couple different arrangements in my mappings config.

LeonHeidelbach:
@siduck it could as well be a random order thing but this issue came up with multiple mappings. And the prune function does not add that much bulk to the utils :wink:

siduck:
> @siduck try moving it before `M.nvimtree` maybe it's not in alphabetical order but just assignment order. I had this issue multiple times and tried a couple different arrangements in my mappings config.

i tried but cant reproduce your issue!

LeonHeidelbach:
@siduck hop on discord and I will show you

siduck (approved):

PR #1085, opened 2022-05-15
fix: remove disabled maps from which-key popup (again)

LeonHeidelbach:
@siduck This is just a small fix becuase disabled bindings showed up in which-key again. Also tested with fresh NvChad installation.

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -64,7 +64,7 @@ nvchad.remove_disabled_mappings = function(key_map)
       for k, v in pairs(key_map) do
          if v ~= nil and v ~= "" then clean_map[k] = v end
       end
-   else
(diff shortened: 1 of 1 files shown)


GreatestOfAllTimee:
did you try without which-key install? im facing the same issue https://github.com/NvChad/NvChad/issues/1071#issuecomment-1126705896

LeonHeidelbach:
> did you try without which-key install? im facing the same issue [#1071 (comment)](https://github.com/NvChad/NvChad/issues/1071#issuecomment-1126705896)

@GreatestOfAllTimee missed this but I fixed it in here https://github.com/NvChad/NvChad/pull/1087. Supporting two different ways of handling key maps is a little cumbersome. Sorry for that.

PR #1095, opened 2022-05-16
fix: disable auto-pairing in javascript named imports

ahhshm:
The problem is that when we use javascript named imports to import a function or method, `nvim-autopairs` adds a pair of parentheses after the function name which we never want. This PR disabled auto-pairing in javascript named imports. 

related: https://github.com/windwp/nvim-autopairs/issues/219


siduck:
@ahhshm i think this is more of a preferability issue so we cant have it ._.

ahhshm:
This is not a preferability issue. We can't call a function in named imports, it's not a valid javascript syntax!!

Issue #1096, opened 2022-05-17
Suggestions on version control


lixvbnet:
I've been using NvChad for a while and am very happy with it. It has really saved me a lot of time from configuring NVIM myself. Thank you guys for contributing your time creating NvChad!

Only one thing I would suggest to improve, that is to version-control all plugins as well as NvChad itself. Whenever users run PackerSync in their setup, it will pull down latest changes of every plugin. Sometimes this is good -- to keep them up-to-date. But it becomes annoying when things break. Even if NvChad never brings breaking changes in the future, plugins might do. For example, I once tried installing v1.0 (MegaChad), which was considered stable at the time, but still got errors when running PackerSync.

We can't control updates of a plugin, but we can choose which version to use. So my idea is, while doing active development consuming latest plugin updates, we can also constantly tag a new "stable build", and it should include current git hashes of all plugins at that time. Users are expected to use latest "stable build" unless they personally choose master branch to take the risk for latest features.

To showcase, I created a script. Here's my suggested workflow:

- Whenever NvChad is relatively stable, run `./build.sh release` to gather current git hashes of plugins to `plugins/versions.lua` ; then commit code changes and create a tag.
- To continue development, run `./build.sh reset` to reset git hashes in `plugins/versions.lua` file to default branch names of the plugins, usually `master` or `main`. Continue delelopment with these changes until next "stable" state.

lixvbnet:
I created a PR to showcase the possible updates.
- After run `./build.sh release` : [plugins/versions.lua](https://github.com/NvChad/NvChad/pull/1097/commits/771c086f97f0534d8dea39855df482f89fc3438f#diff-5d97f7799d49d2592504509620a5843dc381d0a57b736914f05099031d4d2f06) 
- After run `./build.sh reset` : [plugins/versions.lua](https://github.com/NvChad/NvChad/pull/1097/commits/3ce708fd66add5b443a720a7186783b335accb1a#diff-5d97f7799d49d2592504509620a5843dc381d0a57b736914f05099031d4d2f06)

PR #1097, opened 2022-05-17
Suggestions on version control

lixvbnet:
Here's to show case my idea on version control NvChad as well as consumed plugins #1096


siduck:
@lixvbnet this is more user specific and users can just redefine the plugin options in chadrc 

```lua
M.plugins = {
   user = {
      ["NvChad/nvterm"] = {
         commit = "xyz",
      },
   },
}```

lixvbnet:
@siduck Yeah I know users can customize a specific plugin, but my point is when tagging NvChad as a release, it should "snapshot" its dependent plugins as well.

siduck:
@lixvbnet the releases are just to track nvchad config history, arent supposed to be used. You need to use the master config only, release will break when new neovim versions release regardless of plugin's working commit

lixvbnet:
@siduck all right.

PR #1128, opened 2022-05-25
Use `server_capabilities` instead of `resolved_capabilities`

dswij:
`resolved_capabilities` is deprecated in nvim 0.8. All the field in `resolved_capabilities` should also be in `server_capabilities`.

See https://github.com/neovim/neovim/pull/17814


siduck:
you can just replace lspconfig's config for now

PR #1130, opened 2022-05-26
fix: this fixes the `no_whichKey_map` function to now properly work with lspconfig mappings

LeonHeidelbach:
@siduck As the title says, this fixes the problems with the `no_whichKey_map` function and also merges all changes from the test branch. I think this also now completes the lspconfig changes so that we can proceed to merge the dev branch into main.

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -102,7 +102,8 @@ M.comment = {
 M.lspconfig = {
    -- See `<cmd> :help vim.lsp.*` for documentation on any of the below functions
 
-   ignore = { "/lua/plugins/configs/whichkey.lua" },
+   -- define all paths from which these maps should not be applied through the mapping function
+   ignore = { "/lua/custom/init.lua", "/lua/plugins/configs/whichkey.lua" },
 
    n = {
       ["gD"] = {
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -129,7 +129,7 @@ end
 -- register mappings through which-key
 nvchad.whichKey_map = function(maps, opts)
    local present, wk = pcall(require, "which-key")
-   local caller_path = nvchad.get_caller_file_path()
+   local caller_path = nvchad.get_caller_file_path(4)
 
    if not present then
       return false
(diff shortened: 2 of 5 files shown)


siduck:
i'll sleep no so i'll test this tomorrow :sleeping: :rocket:

LeonHeidelbach:
> i'll sleep now so i'll test this tomorrow sleeping rocket

Alright, it should all work now :+1:

siduck:
also one thing! can you make the that function check disabled table directly? those nested 3x+ loops add more complexity and make the code unreadable ;(

LeonHeidelbach:
> also one thing! can you make the that function check disabled table directly? those nested 3x+ loops add more complexity and make the code unreadable kinda ;(

@siduck My man, if you are again talking about the "prune"-function we simply cannot do this. I explained that multiple times now, just checking the disabled key will break functionality. The disabled key is only there to disable maps if you don't want to remap them to anything else. There is just NO OTHER WAY with the current setup of the mapping table to retain this functionality and it is really not that complex. I do not understand why this bothers you so much. If you find another way to retain the same functionality without the loops then all power to you, go ahead and implement it but I do not see any other way. Also I showed you that is has no impact on performace whatsoever so I do not see the issue here.

LeonHeidelbach:
@siduck Oh and btw. it's only 3 loops since I refactored it and there are many other functions that have 3 nested loops :wink:. So it really is not complex at all.

siduck:
@LeonHeidelbach not talking about performance, thats fine. The issue is that those nested loops make the code look unreadable kinda, if you can just check disabled keys in the disabled table only then that would also work right?

https://github.com/NvChad/NvChad/commit/1bc1d9c52af3ad6f2d12ebc68dca2db7fa02dd54#commitcomment-73680058

In this comment you were talking about users disabling their keys outside the disabled table, that'd create more confusion so its better for them to keep it centralized i.e in the disabled table only

LeonHeidelbach:
> @LeonHeidelbach not talking about performance, thats fine. The issue is that those nested loops make the code look unreadable kinda, if you can just check disabled keys in the disabled table only then that would also work right?
> 
> [1bc1d9c#commitcomment-73680058](https://github.com/NvChad/NvChad/commit/1bc1d9c52af3ad6f2d12ebc68dca2db7fa02dd54#commitcomment-73680058)
> 
> In this comment you were talking about users disabling their keys outside the disabled table, that'd create more confusion so its better for them to keep it centralized i.e in the disabled table only

The thing is that in order for people then to be able to properly override key bindings from another plugin that were set by default, they will have to add those bindings to the disabled key first. This will require the user to look into all default bindings first to check whether their new binding has already been used or simply clutter the disabled key by putting all of their binding overrides in it. Without doing so, depending on the load order, either of the bindings could be set randomly which would cause unexpected behaviour. I don't think this is user friendly at all. Also even when only checking the default key, we would still have to keep the loops in order to traverse all keymaps. I have a separate function to create a flat table to create a list of all bindings in their respective modes regardless of the plugin name, that basically creates a huge "disabled" key automatically from all custom user bindings. I do not see why the code seems unreadable to you. If you have any questions on how things work just ask.

PR #1132, opened 2022-05-26
Update statusline.lua

ThatOneOcto:
Changed statusline normal icon to a paper icon, lazy commit but looks better than the original I think


siduck:
@ThatOneOcto this is more about userpreference! so closing this, you can override it in  your config tho

siduck:
https://nvchad.github.io/config/Custom%20config#override-default-config-of-a-plugin

PR #1143, opened 2022-05-29
Remove global vars

### LINKED ISSUE #1142: global var causing issues
So I have a script which reloads init.lua and my plugins_table into neovim whenever the file is written, but since core.utils was replaced with the global `nvchad` variable, the script fails, due to the fact that global variables are not accessible by sourcing or running something like `luafile <file>` or `lua require("<file>")`. when running such commands (important for debugging....)  nvchad will be a blank field and thus throw an error, even though it runs fine on startup.

Was there a particular reason core.utils was changed to be a global variable and have no return value? @siduck It's generally bad practice to use globals because of issues like the above, but if there was something it legitimately addresses I'll make sure that any changes don't cause a regression.

zbirenbaum:
fixes #1142

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -71,7 +71,7 @@ M.bufferline = {
       -- close buffer + hide terminal buffer
       ["<leader>x"] = {
          function()
-            nvchad.close_buffer()
+            require("core.utils").close_buffer()
          end,
          "   close buffer",
       },
--- a/lua/core/options.lua
+++ b/lua/core/options.lua
@@ -1,7 +1,7 @@
 local opt = vim.opt
 local g = vim.g
 
-g.nvchad_theme = nvchad.load_config().ui.theme
+g.nvchad_theme = require("core.utils").load_config().ui.theme
 
 -- use filetype.lua instead of filetype.vim
 g.did_load_filetypes = 0
@@ -86,4 +86,4 @@ vim.schedule(function()
 end)
 
 -- load user options if the file exists
-nvchad.load_config().options.user()
+require("core.utils").load_config().options.user()
--- a/lua/core/packer.lua
+++ b/lua/core/packer.lua
@@ -39,7 +39,7 @@ M.options = {
 }
 
 -- merge overrides if there are any
-M.options = nvchad.load_override(M.options, "wbthomason/packer.nvim")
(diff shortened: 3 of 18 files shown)


siduck:
@zbirenbaum the change has to be made at base46 and extensions repo too so wait up!

siduck:
also i was thinking about moving the core util functions to extensions repo i.e nvchad/init.lua file so those functions would be easily distinguisable in the whole config like 

looks more clear : 
```lua
require("nvchad").load_config()
```

```lua
require("core.utils").load_config()
```

zbirenbaum:
See the command I linked in the issue. You can make this change on as many files as you want at once by running it. Its one line of bash and you can wildcard the whole project

zbirenbaum:
> also i was thinking about moving the core util functions to extensions repo i.e nvchad/init.lua file so those functions would be easily distinguisable in the whole config like
> 
> looks more clear and nvchad specific
> 
> ```lua
> require("nvchad").load_config()
> ```
> 
> ```lua
> require("core.utils").load_config()
> ```

Please don't even consider this. Having core functionality required for running at all should not be spread between repos. Extensions extends NvChad, NvChad doesn't extend extensions. Its in the name.

siduck:
we might change the extensions repo name to utils then @zbirenbaum :laughing:

siduck:
@zbirenbaum  will that bash script change nvchad word to core.utils in comments , variables like ( in vim.g.nvchad_theme)  etc too?

zbirenbaum:
> we might change the extensions repo name to utils then @zbirenbaum laughing

But its not. It also has telescope stuff and the theme switcher etc. These aren't things that need to be loaded by all users. I even keep extensions removed. 

Why would you remove a single small file that every single other file in the project calls to an external repo? Plus it will confuse users who want to call it. This is a really bad idea. Somehow even worse than the global one.

zbirenbaum:
> @zbirenbaum will that bash script change nvchad word to core.utils in comments , variables like ( in vim.g.nvchad_theme) etc too?

It matches `nvchad.` so as long as there is a period after nvchad it will. Otherwise it will not.

siduck:
push on the repo directly

siduck:
@zbirenbaum

PR #1159, opened 2022-05-30
update depreciated functions

AlexanderHOtt:
Neovim warns about `client.resolved_capabilities` and `vim.lsp.buf.formatting` being depreciated.

```
[LSP] Accessing client.resolved_capabilities is deprecated, update your plugins or configuration to access client.server_capabilities instead. 
The new key/value pairs in server_capabilities directly match those defined in the language server protocol
```

```
vim.lsp.buf.formatting is deprecated. Use vim.lsp.buf.format { async = true } instead
```

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -196,7 +196,7 @@ M.lspconfig = {
 
       ["<leader>fm"] = {
          function()
-            vim.lsp.buf.formatting()
+            vim.lsp.buf.format({async=true})
          end,
          "   lsp formatting",
       },
--- a/lua/plugins/configs/lspconfig.lua
+++ b/lua/plugins/configs/lspconfig.lua
@@ -20,8 +20,8 @@ win.default_opts = function(options)
 end
 
 M.on_attach = function(client, bufnr)
-   client.resolved_capabilities.document_formatting = false
-   client.resolved_capabilities.document_range_formatting = false
+   client.server_capabilities.document_formatting = false
+   client.server_capabilities.document_range_formatting = false
 
    local lsp_mappings = utils.load_config().mappings.lspconfig
    utils.load_mappings({ lsp_mappings }, { buffer = bufnr })


lucario387:
Your client.server_capabilities are wrong, or at least it is wrong according to the doc.
The new version should be
```lua
  client.server_capabilities.documentFormattingProvider = false
  client.server_capabilities.documentRangeFormattingProvider = false
```

lua/core/mappings.lua
```diff
@@ -196,7 +196,7 @@ M.lspconfig = {
 
       ["<leader>fm"] = {
          function()
-            vim.lsp.buf.formatting()
+            vim.lsp.buf.format({async=true})
```
LeonHeidelbach, line 199: @AlexanderHOtt `vim.lsp.buf.format` does not exists in NVIM v0.7.0, build type "Release", so this cannot be used for now.

lua/plugins/configs/lspconfig.lua
```diff
@@ -20,8 +20,8 @@ win.default_opts = function(options)
 end
 
 M.on_attach = function(client, bufnr)
-   client.resolved_capabilities.document_formatting = false
-   client.resolved_capabilities.document_range_formatting = false
+   client.server_capabilities.document_formatting = false
```
LeonHeidelbach, line 23: This should be `client.server_capabilities.documentFormattingProvider = false` and `client.server_capabilities.documentRangeFormattingProvider = false`, the field `document_formatting` does not exist in `client.server_capabilities`.

LeonHeidelbach (changes requested):
These changes cannot be merged as they do simply not work. Revert line 199 in "mappings.lua" and correct the server_capabilites.

siduck:
@LeonHeidelbach these are changes are available in 0.8 right? We support 0.7 only, 0.8 users must override some part of the config

LeonHeidelbach:
> This should be client.server_capabilities.documentFormattingProvider = false and client.server_capabilities.documentRangeFormattingProvider = false, the field document_formatting does not exist in client.server_capabilities.

@siduck yes, some of those are only available in 0.8. We can however apply the above changes. The only thing that is not available in 0.7 is `vim.lsp.buf.format`. I have updated my own config to use the new server_capabilities so these could be merged.

GreatestOfAllTimee:
Why not nvchad support for nvim nightly? Just check if version is not 0.7.0

GreatestOfAllTimee:
That's because nvchad can be widely used. nvchad community will grow

GreatestOfAllTimee:
> @GreatestOfAllTimee supporting nightly versions of any software is a pain, since a lot can change between minor version upgrades. In order to provide at least a baseline of reliability in nvim functionality it is always best practice to only support release versions. Supporting nightly would also add useless code for users of other versions. If users decide on using nightly it’s better if they maintain functionality themselves. Same goes for custom forks of the repo.

@LeonHeidelbach 80% agree. I don't know what you mean by useless code. You're already know you can simply check nvim version and also what I mean by _"support neovim nightly"_ only **_depreciated API_** or **_better function handler_** API **NOT** covers the whole new features in nightly. so technically if v0.8+ was released tomorrow, users can simply use that version without waiting days/weeks for nvchad side new upgrades.

LeonHeidelbach:
@GreatestOfAllTimee by useless code I mean for example otherwise unnecessary checks for the version number, that branch into separate logic in order to properly support both versions of neovim. This will add unwanted bloat to the code base and could also make nvchad run slower. Also, if we publicly say that we support nightly versions, we are expected to make every aspect of nvchad work with both versions and handle all issues that come with it. I think @siduck will agree, support for future version prior to their release will be added in the `dev` branch, so that people will not have to wait for too long to upgrade.

GreatestOfAllTimee:
@LeonHeidelbach I don't think just because checking the nvim version will slow down nvchad, give _nanoseconds_ to check the neovim version is acceptable.

```lua
if vim.fn.has('nvim-0.8') == 1 then
    -- (logic v0.8)#v0.8
    client.server_capabilities.documentFormattingProvider = false
    client.server_capabilities.documentRangeFormattingProvider = false
else
    client.resolved_capabilities.document_formatting = false
    client.resolved_capabilities.document_range_formatting = false
end
```

I don't know how vim/lua runtime works, but if you don't call unnecessary logic/scripts for `v0.7.0` than it won't affect performance  **and** if there is no `nvim v0.8+`, it should _ignore all_ logical steps for `#v0.8`

GreatestOfAllTimee:
this is just only my opinions. before this (day 1 of released v.0.7.0) how long nvchad takes to migrate? let's say it only takes 3 days, and after that how many new bugs are introduced?

GreatestOfAllTimee:
> support for future versions prior to their release will be added in the dev branch

make sense.

LeonHeidelbach:
@GreatestOfAllTimee What I said was not necessarily directed at the minor changes in this PR. Nvim v0.7.0 also already supports the new `server_capabilities` field so your `if`-statement would not even be necessary here. My concerns are of a rather general nature. If there are more invasive APIs being deprecated in the future and there is more logic that needs to be adapted in order to fully support the nightly and release versions, this will pollute the code base. There is so much that can go wrong here alongside NvChads own bugs that pop up from time to time, that it is not really worth the headache. At the end of the day it's @siduck 's decision to make, but I would strongly discourage trying to keep up support for nightly and release versions of nvim. The implications this might have on our ability to quickly find and fix potential bugs that occur on nightly but not on the release version and vice versa are too immense imho.

PR #1175, opened 2022-06-06
Organize import for GOLang

waldirborbajr:
The codes is responsible to organize and remove unused imports for GOLang.


bryant-the-coder:
trust me. siduck will say add this in ur userconfig

siduck:
> trust me. siduck will say add this in ur userconfig

yes because this is more of a user specific thing so we dont add stuffs that are bound to or favor a specific language etc, whats wrong in this ?

siduck:
@waldirborbajr its better to add it here https://github.com/NvChad/nvchad.github.io/blob/src/docs/config/Tricks.md

PR #1189, opened 2022-06-10
Adding configuration for file_info path statusline

Freyskeyd:
I was facing an annoying issue lately where I was editing a lot of files with the same name.
I updated my configuration to have the opportunity to define a full path instead of just a file name.

Here's what I did, I'm open to any suggestion to improve that kind of configuration.
I didn't want to go to a fully customizable statusline because the PR would have been much longer, and I only had an issue with FileInfo, but feel free to suggest changes for future usages.
 
Signed-off-by: Freyskeyd <simon.paitrault@gmail.com>

### DIFF
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -35,6 +35,7 @@ M.plugins = {
       },
       statusline = {
          separator_style = "default", -- default/round/block
+         file_info_style = "default", -- default/full_path
          config = "%!v:lua.require'ui.statusline'.run()",
       },
    },
--- a/lua/ui/statusline.lua
+++ b/lua/ui/statusline.lua
@@ -22,6 +22,20 @@ local sep_style = {
    },
 }
 
+local file_info_styles = {
+  default = function()
+    return fn.fnamemodify(fn.expand "%:t", ":r")
+  end,
+
+  full_path = function ()
+    return fn.expand "%"
+  end
+}
+
+local user_file_info_style = require("core.utils").load_config().plugins.options.statusline.file_info_style
+local file_info_style = file_info_styles[user_file_info_style]
+
+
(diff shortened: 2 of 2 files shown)


tomasky:
As far as I know, If some files have the same name，the  [bufferline.nvim](https://github.com/akinsho/bufferline.nvim) will show full path of these files

siduck:
> As far as I know, If some files have the same name，the [bufferline.nvim](https://github.com/akinsho/bufferline.nvim) will show full path of these files

agreed

siduck:
@Freyskeyd really appreciate the PR but bufferline already does the job so Ig we dont have to add this in the statusline. I'll soon make statusline overridable so you can just edit it via chadrc and add your own stuff

Freyskeyd:
@siduck @tomasky Oh ok, I don't use bufferline, that's maybe why I didn't see that :)

siduck:
> @siduck @tomasky Oh ok, I don't use bufferline, that's maybe why I didn't see that :)

no worries, I'll definitely add options to override the statusline soon. for now there's a temporary fix that you can try : 

image

change this path to your statusline's function. I'll have to rewrite the nvchad website for various reasons so making statusline overridable might take a while!

cgxxv:
So what is the right way to show the file path info in the statusline? Did `NvChad` support this feature right now?

tomasky:
My ubuntu terminal title doesn't show the file path now, I dont know why. @siduck ,Is yours showing 　 normally?

siduck:
@tomasky `vim.opt.title = true`, some users with specific terminal had this issue. its a cmp issue with this option on, so i removed it

tomasky:
ok ,it works ,thanks

PR #1202, opened 2022-06-13
replaced nvim-gps with nvim-navic

sanchay9:
Simply replaced [nvim-gps](https://github.com/SmiteshP/nvim-gps) with the improved version [nvim-navic](https://github.com/SmiteshP/nvim-navic)


lucario387:
This is missing a lot of navic stuffs
- First, icon color groups are no longer in the form of `["smth-name"]`, but rather just `Smth`.
- Next, nvim-navic requires it to be attached to an LSP server so that it can get the current context. So where is the change to `on_attach` in lspconfig?
- Changing the hlgroup from `Gps...` to `Navic...` makes it so default base46 no longer provide any highlight, have you also created a PR in base46 for changing such hlgroups?

lua/plugins/configs/navic.lua
```diff
@@ -0,0 +1,39 @@
...
+   return
+end
+
+local icons = require "plugins.configs.lspkind_icons"
+
+local options = {
+   depth = 5,
+
+   icons = {
+      ["class-name"] = "%#NavicItemKindClass#" .. icons.Class .. "%*" .. " ",
```
lucario387, line 13: For example, `["class-name"]` should be changed to `Class` (according to Navic doc)

siduck:
@sanchay9 no not yet, I will add nvim-navic when 0.8v releases (in winbar probably) . So i will soon remove gps

siduck:
i did try navic yesterday and the icons werent coloring properly so i'll keep that on hold

cj:
@siduck Is there a reason nvim gps was removed? I loved that feature!

cj:
@siduck What do you think about the idea of adding hooks to the status line, so that you can add custom things to it without touching the main NvChad files?

siduck:
> @siduck What do you think about the idea of adding hooks to the status line, so that you can add custom things to it without touching the main NvChad files?

I will make it overridable soon so wait for a while

siduck:
> @siduck Is there a reason nvim gps was removed? I loved that feature!

gps is depreceated and i couldnt configure navic icons colors well so i kept that on hold until v0.8 nvim releases, we'll add it to winbar maybe

Issue #1201, opened 2022-06-13
chadrc.lua not applied until override.lua is manually changed


jrozhon:
**Describe the bug**
I am trying to automate nvim installation. Whenever I issue the following commands to make "a clean install" of nvchad, my lua/custom/chadrc.lua configuration is ignored until I manually modify the file or the file it links to.

**To Reproduce**
Clean install by this sequence:
```
rm -rf ~/.config/nvim
rm -rf ~/.local/share/nvim
rm -rf ~/.cache/nvim
git clone https://github.com/NvChad/NvChad ~/.config/nvim --depth 1; nvim
ln -s ~/dotfiles/nvim/custom ~/.config/nvim/lua/custom
touch -a -m ~/dotfiles/nvim/custom/override.lua
```
Then when I run nvim, the configuration from chadrc.lua is not applied and custom modules are not installed.

**Expected behavior**
A complete working configuration.

**Screenshots**
See video in attachment.

**Desktop (please complete the following information):**
 - Ubuntu 22.04
 - Alacritty 0.11
 - Neovim 0.8.0-dev

**Additional context**

https://user-images.githubusercontent.com/12800857/173343982-98188f14-7701-45ce-b291-07daf1085a0c.mp4

tomasky:
try this:

```shell
rm -rf ~/.config/nvim
rm -rf ~/.local/share/nvim
rm -rf ~/.cache/nvim
git clone https://github.com/NvChad/NvChad ~/.config/nvim --depth 1
ln -s ~/dotfiles/nvim/custom ~/.config/nvim/lua/custom
touch -a -m ~/dotfiles/nvim/custom/override.lua
nvim
```

tomasky:
Why this, maybe this, i dont known ,you can remove this to test it
```bash
touch -a -m ~/dotfiles/nvim/custom/override.lua
```

jrozhon:
Yea, this is only to "simulate" the file change, but this does not work either. I literally have to save the chadrc.lua or override.lua for the changes to take effect.

tomasky:
i canot reproduce this ,my custom plugins are installed correctly without `PackerSync`,maybe you can provide your configs,then i can test it.

jrozhon:
Thank you for your effort. Configs are attached. No changes to nvchad core were made
[custom.tar.gz](https://github.com/NvChad/NvChad/files/8892014/custom.tar.gz)
.

tomasky:
[...]

PR #1213, opened 2022-06-15
feature: make override can be a function

a690700752:
so we can use `require` even though it been lazy load like this.
```lua
local function override_cmp()
  local cmp = require 'cmp'
  return {
    mapping = {
      ['<C-k>'] = cmp.mapping.select_prev_item(),
      ['<C-j>'] = cmp.mapping.select_next_item(),
      ["<Tab>"] = cmp.mapping(function(fallback)
        local copilot_keys = vim.fn['copilot#Accept']()
        if copilot_keys ~= '' and type(copilot_keys) == 'string' then
          vim.api.nvim_feedkeys(copilot_keys, 'i', true)
        elseif cmp.visible() then
          cmp.select_next_item()
        elseif require("luasnip").expand_or_jumpable() then
          vim.fn.feedkeys(vim.api.nvim_replace_termcodes("<Plug>luasnip-expand-or-jump", true, true, true), "")
        else
          fallback()
        end
      end, {
         "i",
         "s",
      }),
    }
  }
end

M.override = {
  ["hrsh7th/nvim-cmp"] = override_cmp,
}

```
#1201

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -148,6 +148,10 @@ end
 M.load_override = function(default_table, plugin_name)
    local user_table = M.load_config().plugins.override[plugin_name]
 
+   if type(user_table) == "function" then
+      user_table = user_table()
+   end
+
    if type(user_table) == "table" then
       default_table = merge_tb("force", default_table, user_table)
    else


gmartsenkov:
Brilliant 🎉

PR #1220, opened 2022-06-16
Add mapping for new floating term

TianYaX:
Alt key is not friendly in macOS

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -308,6 +308,13 @@ M.nvterm = {
 
       -- new
 
+      ["<leader>i"] = {
+         function()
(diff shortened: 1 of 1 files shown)


strix:
FWIW I think alt being "not friendly in macOS" is mostly a terminal emulator problem. On most terminal emulators you can send the option key as alt. I'm using kitty with `macos_option_as_alt yes` set and the default alt mappings work for me. I know there are similar settings in the default Terminal.app and iTerm2.

siduck:
i agree with strix, this is more of a macos terminal related issue so cant help. U can just override your those alt mappings if u are unable to make alt work on macos

Issue #1224, opened 2022-06-17
Keep the cursor position while selecting a word (asterisk)


horta:
Enter the following in your neovim:

```
nnoremap * <Cmd>let @/='\<'.expand('<cword>').'\>'<bar>set hlsearch<CR>
```

Typing * now selects the words without moving the cursor. I bet you will find it very useful =)

Source: https://vi.stackexchange.com/a/22889

siduck:
@horta can you provide a video of how this looks? I'll add it in the main repo if I like it :D

horta:
https://user-images.githubusercontent.com/514522/174630722-36eb2ed1-4790-4d1d-8e5f-0d74dba1f38a.mov

I type `*` to search BUFFSIZE (word under the cursor). Then I enable the above proposal and type `*` for  BUFFSIZE and other words. Notice that the cursor select the word under the cursor (and highlight all of them in the screen) without moving the screen.

It is very nice =)

LeonHeidelbach:
@siduck @horta yes, we definitely need this, the default jumping behavior is so annoying!

horta:
That is such a nice feature, right? It basically replaces plugins that highlight words under the cursor and maintain the same functionality that `*` has. I will have a look on how to make it in lua, @siduck Thanks!

siduck:
yep I love it, add a PR for it :)

horta:
Can I ask questions here before I open an PR?

I managed to translate it to lua:

```lua
vim.fn.setreg("/", "\\<" .. vim.fn.expand("<cword>") .. "\\>")
vim.opt.hlsearch = true
```

The first line set the register `"/` to the string `\<word_under_the_cursor\>` and then set the `hlsearch` option to true.

Would it make sense to have a lua function like

```lua
function Shine()
    vim.fn.setreg("/", "\\<" .. vim.fn.expand("<cword>") .. "\\>")
    vim.opt.hlsearch = true
end
```

?


I'm aware that Nvchad use plugins developed by @siduck , does it make sense to have that function in there?

siduck:
@horta add it in the core.options.lua file

PR #1253, opened 2022-06-21
Keep the cursor position while pressing *.

horta:
Implement the feature discussed in https://github.com/NvChad/NvChad/issues/1224

### DIFF
--- a/lua/core/options.lua
+++ b/lua/core/options.lua
@@ -99,5 +99,14 @@ vim.schedule(function()
    vim.cmd [[ silent! rsh ]]
 end)
 
+-- Keep the cursor position while forward-searching a word.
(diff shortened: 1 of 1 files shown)


siduck:
as @horta seems dead, i will close this PR and we will prefer the default behaviour as pressing `*` cycles over the highlighted keywords which is nice imo but users can still keep this in their custom mappings 

```lua
["*"] = {
      function()
        vim.fn.setreg("/", "\\<" .. vim.fn.expand "<cword>" .. "\\>")
        vim.opt.hlsearch = true
      end,
      "kek",
    },
```

horta:
Not dead ;) Just didn't know what else to do...

But if the feature is not popular, better no have it.

Btw, I have been using https://github.com/RRethy/vim-illuminate for similar effect. It is quite handy one.

PR #1259, opened 2022-06-23
Minor syntax fix

Titor8115:
I was messing around with configurations of the cmp.nvim plugin. And it seems that the mapping for Confirm won't actually use the given value of `select=` until I put parenthesis around blocks.
I'm quite new to Lua script so I'm not entirely sure if this is intentional or not.


siduck:
@Titor8115 thats from the default config of cmp 

image

Titor8115:
Nvm... It seems that I still haven't wrap my head around Lua yet.
Tx for the great project.
Closing the Pull request

PR #1270, opened 2022-06-26
feat(nvim-tree): enable adaptize size

hown3d:
Signed-off-by: Höhl, Lukas <lukas.hoehl@accso.de>

Nvim-tree introduced an option to dynamically increase and shrink the size of the drawn tree.  \
See https://github.com/kyazdani42/nvim-tree.lua/pull/1293.

I think this is something, that nearly everyone wants in their default config, which is why this PR enables this feature in the default config of nvim-tree.


siduck:
@hown3d can you show a video of with / without this? Im confused

PR #1328, opened 2022-07-11
fix key mappings organized into groups

P1P0S:
whichkey organized into its proper groups

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -59,27 +59,26 @@ M.general = {
 }
 
 M.tabufline = {
+  n = {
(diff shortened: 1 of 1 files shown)


the-eduardo:
image
Line 107 - Duplicated line 
`nvim/lua/core/mappings.lua:107: '}' expected (to close '{' at line 105) near '='
stack traceback:`

siduck:
@the-eduardo  really apprecitae the PR but this is just a matter of preference and those missing groups arent much of an issue, we will add future mappings to that group and this will be a breaking change as many core mappings change so closing this

PR #1370, opened 2022-07-25
fix: remove duplicated item from 'lsp_cmds'

ahhshm:


siduck:
@ahhshm i'll be removing lspinstaller and replace it with mason.nvim so this PR would not be helpful ig!

siduck:
https://user-images.githubusercontent.com/59060246/180693605-35e51db6-30a6-435d-a49a-4b43a30df332.mp4


Only issue i face right now is that lazyloading lspconfig will not get the lsp server attach on buffer automatically ( it starts in insert mode or lspStart/LspRestart )

PR #1379, opened 2022-07-26
fix(0.8): :bug: update session data dir in neovim nightly

tomasky:
Nvim now stores "session data" (shada, persistent undo, ...) in $XDG_STATE_HOME (~/.local/state)

### DIFF
--- a/lua/core/options.lua
+++ b/lua/core/options.lua
@@ -14,7 +14,6 @@ if g.vim_version < 8 then
   g.do_filetype_lua = 1
 end
 
-
 opt.laststatus = 3 -- global statusline
 opt.showmode = false
 
@@ -105,6 +104,6 @@ end
 
 -- set shada path
 vim.schedule(function()
-  vim.opt.shadafile = vim.fn.expand "$HOME" .. "/.local/share/nvim/shada/main.shada"
+  vim.opt.shadafile = vim.fn.stdpath(g.vim_version > 7 and "state" or "data") .. "/shada/main.shada"
   vim.cmd [[ silent! rsh ]]
 end)


lua/core/init.lua
```diff
@@ -1,6 +1,9 @@
 -- add binaries installed by mason.nvim to path
 local is_windows = vim.loop.os_uname().sysname == "Windows_NT"
-vim.env.PATH = vim.env.PATH .. (is_windows and ";" or ":") .. vim.fn.stdpath "data" .. "/mason/bin"
+vim.env.PATH = vim.env.PATH
+  .. (is_windows and ";" or ":")
+  .. vim.fn.stdpath(vim.g.vim_version > 7 and "state" or "data")
+  .. "/mason/bin"
```
lucario387, line 6: This is to add the path of `mason.nvim` to the PATH env, and by default, it is always added to `vim.fn.stdpath("data")`, so no need to check for it here

lua/core/packer.lua
```diff
@@ -2,7 +2,7 @@ local M = {}
 
 M.bootstrap = function()
   local fn = vim.fn
-  local install_path = fn.stdpath "data" .. "/site/pack/packer/start/packer.nvim"
+  local install_path = fn.stdpath(vim.g.vim_version > 7 and "state" or "data") .. "/site/pack/packer/start/packer.nvim"
```
lucario387, line 5: Similarly here, `packer.nvim`, if you are using the git command/Use NvChad as a Packer Bootstrap, then packer will only be downloaded at stdpath("data"). No need for an extra check here

lua/core/init.lua
lucario387: Not to mention doing this renders 0.8.0 users complete broken installs for packages downloaded by mason, as you have to use absolute paths to get the packages, re-add the binary path to `$PATH`, or change default install directory of mason to the path you set here

lucario387 (changes requested):

siduck:
@tomasky ready to merge?

tomasky:
yes

PR #1395, opened 2022-07-28
feat(keymaps): add keymaps to move in the wildmenu

ahhshm:
These keymaps select the next/previous item in the wildmenu with `C-j` and `C-k`. I think these are more intuitive than `C-n` and `C-p` (default keymaps) and kinda in line with the keymaps that already exist (`C-j/k` is also used to move between windows and navigate within insert mode).


siduck:
@ahhshm 

image


It creates these list of dots when I cycle through

siduck:
And ig its better to just keep it default as we use Tab / Shift Tab or C-n / C-p to cycle through a menu in nvim , even cmp , telescope and so on. So closing this!

PR #1403, opened 2022-07-29
Register mappings only if pluigin is enabled | Misc

Akianonymus:
Use setup of packer to load the mappings before

Load cmp on cmp module

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -8,7 +8,6 @@ local M = {}
 
 M.general = {
   i = {
-
     -- go to  beginning and end
     ["<C-b>"] = { "<ESC>^i", "論 beginning of line" },
     ["<C-e>"] = { "<End>", "壟 end of line" },
@@ -21,7 +20,6 @@ M.general = {
   },
 
   n = {
-
     ["<ESC>"] = { "<cmd> noh <CR>", "  no highlight" },
 
     -- switch between windows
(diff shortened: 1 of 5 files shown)


siduck:
@Akianonymus wait for this, i was thinking to remove whichkey ( will do a POLL )  and load mappings of all plugins in their setups, like we do for lspconfig 

```lua
  local lsp_mappings = utils.load_config().mappings.lspconfig
  utils.load_mappings({ lsp_mappings })
```

So i'd load the one in tabufline config too, so if its disabled then mappings will be too

Akianonymus:
Ok, but you cannot move mappings to other place, it needs to be in packer setup block, because we need to load the mappings before the plugins are loaded, are we are lazy loading the plugins on the mappings ( most of it atleast ).

siduck:
> Ok, but you cannot move mappings to other place, it needs to be in packer setup block, because we need to load the mappings before the plugins are loaded, are we are lazy loading the plugins on the mappings ( most of it atleast ).

yes it'll go in the setup part and we're currently not lazyloading the mappings, will do soon

Akianonymus:
> yes it'll go in the setup part and we're currently not lazyloading the mappings, will do soon

Ok, ping when done.

Akianonymus:
@siduck When merge.

siduck (approved):

PR #1444, opened 2022-08-08
'blankline' fix filetype_exclude mason

nulloneguy:
with a capital letter it doesn't work

screenshot


siduck:
@nulloneguy 
image


same issue :/ 

open a file then open mason

PR #1489, opened 2022-08-19
Resize panes when vim window is resized

SebastienElet:
I always use nvim in tmux and I often change my panes size.

# Before

CleanShot 2022-08-19 at 18 17 04

# After

CleanShot 2022-08-19 at 18 21 10


siduck:
@SebastienElet didnt know this was possible, thank you so much!

PR #1493, opened 2022-08-20
fix: Windows User cannot load gitsigns

lucario387:
As Windows Syntax for Pipe is different, remove Pipe completely to make sure Gitsigns can be loaded when using on Windows Powershell


siduck:
would this affect non windows OS's?

lucario387:
> would this affect non windows OS's?

@siduck no, there will be no changes for non windows OS's.   
The pipe part (`2>/dev/null`) is only for Linux and Mac, so it was making the command always returning errors (as it is simply), leading to gitsigns never loading.   
For Linux and Mac users, there will be no changes in functionality due to this change

Issue #1518, opened 2022-08-28
NEED YOUR SUGGESTION | Proposal to simplify plugin management |


musjj:
**Is your feature request related to a problem? Please describe.**
I feel that the current plugin management framework can be greatly simplified and thus be more flexible. Right now, we need to handle 3 separate tables just to manage and configure our plugins: `user`, `override` and `remove`. I find this unnecessarily complicated, awkward and limiting, when there's a much simpler way that would give the user a lot more flexibility and control.

**Describe the solution you'd like**

`plugins/configs/plugin.lua`
First the plugin configuration should return the options table to be passed to `setup()`:
```lua
vim.opt.opt_foo = true

local options = {
  opt_bar = true,
}

return options
```

`plugins/init.lua`
Then, we move the `setup()` call to the packer `config` recipe. We then can pass the whole plugin table to the user's config and let them do whatever they want with the table:
```lua
vim.cmd "packadd packer.nvim"

local plugins = {
  ["author/plugin.nvim"] = {
    config = function()
      local options = require "plugins.config.plugin"
      require("plugin").setup(options)
    end,
  },
}

local present, custom = pcall(require, "custom.plugins")

if present then
  plugins = custom(plugins)
else

require("core.packer").run(plugins)
```

`custom/plugins.lua`
In the user configuration, we can now manipulate the plugins table with great freedom:
```lua
return function(plugins)
  -- Modify options table
  plugins["author/plugin.nvim"].config = function()
    local options = require "plugins.config.plugin"
    options.opt_bar = false
    require("plugin").setup(options)
  end

  -- Modify repository URL
  plugins["fork/plugin.nvim"] = plugins["author/plugin.nvim"]

  -- Remove plugin
  plugins["author/plugin.nvim"] = nil

  -- Add our own plugin
  plugins["our/own.plugin"] = {}

  return plugins
end
```

What do you think? I feel that this method of configuration is a lot nicer that what we currently have. Any suggestions for improvements are welcome.

siduck:
the current implementation is already very simple :l

siduck:
please read this section well https://nvchad.com/config/plugins

musjj:
Have you read my suggestions? Please have some considerations before hurrying to close it.
The current implementation is simple, but it still adds unnecessary abstractions that makes plugin management awkward and unnatural.
[...]

PR #1531, opened 2022-09-02
Simplify plugin management even more (#1518)

siduck:
## Things that are new with the syntax : 

1. You can now override plugin configs without entering chadrc, just put it in `override_options` of your plugin table.

2. To remove a plugin, just set it to false 

## Examples : 

```lua
-- chadrc.lua
M.plugins = require "custom.plugins"
```

```lua
-- custom/plugins.lua
return {
  -- Install a plugin
  ["Pocco81/TrueZen.nvim"] = {},

  -- Override packer definition options of default plugin
  ["neovim/nvim-lspconfig"] = {
    event = "VimEnter",
  },

  -- Override options from plugin config
  ["kyazdani42/nvim-tree.lua"] = {
    override_options = {
      view = {
        side = "right",
        width = 25,
      },
    },
  },

  -- remove default plugin , set it to false
  ["lewis6991/gitsigns.nvim"] = false,
}
```

The above is same as this : 

```lua
-- chadrc.lua
M.plugins = {
 
   ["Pocco81/TrueZen.nvim"] = {},
   ....
}
```


tomasky:
@siduck  Is  the  custom/plugins.lua file  allowed to  require lua module?

siduck:
> @siduck Is the custom/plugins.lua file allowed to require lua module?

yes  

```lua
  ["hrsh7th/nvim-cmp"] = {
    override_options = function()
      local cmp = require "cmp"

      return {
        mapping = {
          ["<C-d>"] = cmp.mapping.scroll_docs(-8),
        },
      }
    end,
  },

```

tomasky:
no ,i mean ,
```lua
local configs = require "custom.configs"
["hrsh7th/nvim-cmp"] = {
    override_options = configs.cmp
  },

```

siduck:
> no ,i mean ,
> 
> ```lua
> local configs = require "custom.configs"
> ["hrsh7th/nvim-cmp"] = {
>     override_options = configs.cmp
>   },
> ```

yes this will work

Akianonymus:
Also, what happened to all configs returning the table ? I don't see that change

Akianonymus:
@siduck Yes, did you forget, it was being discussed in that issue.

siduck:
With the new syntax of overriding plugin configs, we no longer need to import plugin configs so we don't have to export them ( return their tables )

Akianonymus:
> With the new syntax of overriding plugin configs, we no longer need to import plugin configs so we don't have to export them ( return their tables )

Returning tables so user can fetch and do whatever they want with it, it wasn't because it was needed with the new syntax.

siduck:
What will the user do apart from overriding the plugin config?

siduck:
> Sometimes, i want a part of the config, like capabilities part of lspconfig, etc and use it somewhere. Yes, one could just always copy paste but that's the solution to everything anyway, lol.
> 
> Also, it's more about improving the config structure. Each plugin should return a table, containing two things, a setup function, and a config table. I am just trying to remove the unevenness of the config.

capabilities in lspconfig is already returned, you already use it in your custom config . Anything other edge cases?

Akianonymus:
> also its not un-evenness, even if you look out into non nvchad configs ( general nvim configs ) the whole plugin config + its setup call is in the same file and not separated into functions

We will talk about it later, anyway, first let't merge this first, i will test it soon.

siduck:
i tested it here already but test it from dev branch directly first then I'll merge this 

```lua
M.options = {
	nvChad = {
		update_branch = "dev",
	},
}
```

Akianonymus:
Your dev branch is outdated btw, doesn't have the latest commits of master ( therefore breaks colorizer ).

ALso i got this error

`
packer.nvim: Error running config for ui: [string "..."]:0: attempt to call field 'nvchad_ui' (a nil value)
`

Akianonymus:
Nevermind, just colorizer is broken, rest fine, just pick the latest commits from master. Colorizer will work fine too i guess.

Rest of the stuff will check tonight ( new override method )

Akianonymus:
No changes for now, just pulled the latest commit and force pushed.

siduck:
dang soo much extra code! waiting for next month to clean it all :angry: :rocket:

Akianonymus:
@siduck 😂 Btw, do pin a issue, so users can change, with the link to the guide.

PR #1542, opened 2022-09-07
Fix colorizer

Akianonymus:
…ds case insensitive

just as neovim works


Akianonymus:
@siduck  Test and report for the keymap stuff

PR #1543, opened 2022-09-07
Make luasnip support vscode,snipmate,lua three types of loader.

kvzn:
Reference: https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#loaders

If this PR is accepted, the document will need to be updated:
```lua
-- specify the full path or relative to the directory of $MYVIMRC
-- vscode format
vim.g.vscode_snippets_path = 
-- snipmate format
vim.g.snipmate_snippets_path =
-- lua format
vim.g.lua_snippets_path =
```


modsyan:
thx alot ❤️

siduck:
@kvzn only make the change of luasnip config

siduck:
@kvzn also why calling the load function 3 times? cant it be called only once

PR #1547, opened 2022-09-08
Support plugin removal using old syntax until remove

lucario387:
In current setup, plugins defined in `chadrc.plugins.remove` are not being called, so it's not removed.   
Add back the remove using that table


siduck:
i dont think the change to colorizer config had to be done, the PR had no merge conflicts with the main branch so colorizier conf from main branch would still remain the same

PR #1601, opened 2022-09-26
feat: add buffer mapping for gitsigns

jtuz:
Features Added:
- Buffer mappings for gitsigns

Reasoning:
The default settings from [ewis6991/gitsigns.nvim](https://github.com/lewis6991/gitsigns.nvim#keymaps) does not include a default mapping for all gitsigns actions, also NvChad does not include this as well.

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -404,4 +404,137 @@ M.blankline = {
   },
 }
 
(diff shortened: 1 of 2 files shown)


siduck:
@jtuz send the PR in the dev branch.

PR #1620, opened 2022-10-07
fix(options): since this is 0.8, there is no need for this option

bryant-the-coder:


siduck:
@bryant-the-coder already made that change into dev branch!

PR #1632, opened 2022-10-11
Add support for Up and Down arrow keys in cmp

tabotkevin:
The current cmp configuration options does not have support for Up and Down arrow key during selection in auto-complete which is an issue for users who are used to using the Up and Down arrow keys.

It will be nice to have support Up and Down arrow keys during auto-complete selections.


siduck:
@tabot-kevin users can add this themselves

```lua

  ["hrsh7th/nvim-cmp"] = {
    override_options = function()
      local cmp = require "cmp"

      return {
        mapping = {
          ["<Up>"] = cmp.mapping.select_prev_item(),
          ["<Down>"] = cmp.mapping.select_next_item(),
        },
      }
    end,
  },
  ```

PR #1644, opened 2022-10-21
feat: allow overwrite alpha buttons

jtuz:
Why:

Alpha dashboard provides a cool way to define shortcuts to start using neovim,this feature allows to NvChad overwrite mappings according user needs in the initial dashboard.

This solve the needed from this thread: https://t.me/DE_WM/542552

Just right now the default mapping from NvChad has some conflicts, for example with the bookmarks ` <Leader>bm`  is omitted in favor of  ` <Leader>b` for new buffer, and marks are a good way to set a bookmark for large project and big modules, so this PR allow to override the dashboard buttons just by requiring ` local buttons = require("core.utils").alpha_button`  in the user namespace that allow overwrite default NvChad settings for Alpha dashboard

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -189,4 +189,32 @@ M.packer_sync = function(...)
   end
 end
 
+M.alpha_button = function (sc, txt, keybind)
+  local sc_ = sc:gsub("%s", ""):gsub("SPC", "<leader>")
+
+  local opts = {
+    position = "center",
+    text = txt,
+    shortcut = sc,
+    cursor = 5,
+    width = 36,
+    align_shortcut = "right",
+    hl = "AlphaButtons",
+  }
+
+  if keybind then
+    opts.keymap = { "n", sc_, keybind, { noremap = true, silent = true } }
+  end
+
+  return {
+    type = "button",
+    val = txt,
+    on_press = function()
+      local key = vim.api.nvim_replace_termcodes(sc_, true, false, true) or ""
+      vim.api.nvim_feedkeys(key, "normal", false)
+    end,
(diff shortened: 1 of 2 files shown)


lua/core/utils.lua
```diff
@@ -189,4 +189,32 @@ M.packer_sync = function(...)
   end
 end
 
+M.alpha_button = function (sc, txt, keybind)
+  local sc_ = sc:gsub("%s", ""):gsub("SPC", "<leader>")
```
lucario387, line 193: Would be better if you also change it to show leader key, instead of defaulting to the showtext that is SPC.
There are people who set their leader to be not space

jtuz: I think this is the best approach for now, a great power comes with a big responsibility, so the tradeoff for this flexibility to overwrite the alpha dashboard put to users the responsibility to map this according their needs, you can watch what is coming with this PR whit this little proof of concept:

```lua
local function define_map(sc)
  return sc:gsub("%s", ""):gsub("SPC", "<leader>")
end


local lunit = require "lunitx"

if _VERSION >= 'Lua 5.2' then
    _ENV = lunit.module('simple','seeall')
else
    module( "simple", package.seeall, lunit.testcase )
end

function test_with_SPC()
  assert_equal(define_map("SPC a b"), "<leader>ab")
end

function test_with_leader()
  assert_equal(define_map"<leader> a b", "<leader>ab")
end

function test_normal_mapping_composed_keys()
  assert_equal(define_map("a b"), "ab")
end

function test_normal_mapping_single_key()
  assert_equal(define_map("a"), "a")
end

```

PR #1695, opened 2022-11-24
Fix next diagnostic key mapping

Crescent617:
Consistent with vim's convention, may be this key mapping should be `]d`?


siduck:
@Crescent617 its already pushed in the dev branch

siduck:
will be included in v2.0

Issue #1690, opened 2022-11-22
custom chadrc errors/swallowed masked


mattbdc:
Getting to grips with nvchad now however spent a good few hours struggling blindly with a failing custom.chadrc - should the error get checked in core.utils  (no else branch on the error check) - how about if error ends with "not found" then remain silent - otherwise print the error (and terminate) ?  silence seems the wrong choice here but I must admit I'm new here. I temporarily replaced the pcall wrappred require custom.chadrc in core.utils here so I fail fast but obvious one can be smarter here for the expected non-exceptional case where custom.chadrc isn't found

I'm wondering if this is desirable before proceeding - I realize one could potentially test if this is failing by requiring via :lua require("custom.chadrc") via the command mode in a nvim instance but I think that's several hours down the line from hour 0 - and people are likely to tweak in custom before hour 7 imo..  are we optimizing for performance at the cost of approach-ability ?

siduck:
i will soon remove the things that were preventing errors to show up, example https://github.com/NvChad/NvChad/blob/main/lua/core/utils.lua#L12 

@Akianonymus had added it previously

mattbdc:
https://github.com/NvChad/NvChad/blob/main/lua/core/utils.lua#L16 you might be able to catch some config mistakes intelligently but you won't catch everything as you have no else branch here

what if there's a typo, a bad variable, a print(nonexist + 1) just to illustrate

mattbdc:
```
if chadrc_exists then
    -- merge user config if it exists and is a table; otherwise display an error
    if type(chadrc) == "table" then
      config.mappings = M.remove_disabled_keys(chadrc.mappings, config.mappings)
      config = merge_tb("force", config, chadrc) or {}
    else
      error "chadrc must return a table!"
    end
 else
   if not chadrc.endswith("not found:") then
    print("loading custom.chadrc threw an error - check your custom.chadrc and any downstream requires")
     print(chadrc)
   end 
 end
```

i dont know if endswith existss

siduck:
@mattbdc we cant print the "chadrc not found" error everytime, because there are users who just want to use vanilla nvchad i.e without chadrc etc then according to your code, they will get that "chadrc not found" warning every time which isnt good.

mattbdc:
Note the NOT, if it isn't found the error message ends with not found and thus we remain silent

siduck:
hmm i dont think there's an endwith function in lua

siduck:
[...]

PR #1704, opened 2022-12-06
ensure errors arent silently swallowed in custom chadrc if exists

mattbdc:
Avoids using pcall as a general catch all so that errors in one's custom chadrc aren't silently swallowed. 

Addresses issue #1690

Due credit to @ii14 for approach here..

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -3,9 +3,12 @@ local merge_tb = vim.tbl_deep_extend
 
 M.load_config = function()
   local config = require "core.default_config"
-  local chadrc_exists, chadrc = pcall(require, "custom.chadrc")
+
+  -- nvim_get_runtime_file always uses forward slashes regardless of platform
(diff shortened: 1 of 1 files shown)


lua/core/utils.lua
```diff
@@ -3,9 +3,12 @@ local merge_tb = vim.tbl_deep_extend
 
 M.load_config = function()
   local config = require "core.default_config"
-  local chadrc_exists, chadrc = pcall(require, "custom.chadrc")
+
+  -- nvim_get_runtime_file always uses forward slashes regardless of platform
+  local chadrc_exists = vim.api.nvim_get_runtime_file('lua/custom/chadrc.lua', false)[1]
 
   if chadrc_exists then
+    local chadrc = require  "custom.chadrc"
```
ii14, line 11: Try something like this instead
```suggestion
  local chadrc_path = vim.api.nvim_get_runtime_file('lua/custom/chadrc.lua', false)[1]

  if chadrc_path then
    local chadrc = dofile(chadrc_path)
```

siduck (approved):

PR #1729, opened 2022-12-31
remove PackerSync VimEnter autocmd

lucario387:
Don't see why we need to defer it till VimEnter, so just remove it and wrap the completion function instead (so that it won't start packer right away, maybe?)


siduck:
@lucario387 this needs to be put on hold, i was thinking about replacing packer with lazy.nvim in v2.0 so we might not need these autocmds etc

PR #1744, opened 2023-01-16
fix wrong nvim-tree.lua reporitory name error

anosillus:
Now packer.nvim  is outputting error in NvChad at startup from Packer.nvim.
Because the repository `kyazdani42/nvim-tree.lua` that NvChad is referencing doesn't exist.

The repository moved to `nvim-tree/nvim-tree.lua`
[nvim-tree/nvim-tree.lua: A file explorer tree for neovim written in lua](https://github.com/nvim-tree/nvim-tree.lua)

So I have updated `kyazdani42/nvim-tree.lua` to `nvim-tree/nvim-tree.lua` in two files.
This change fixes packer.nvim.
I confirmed that the error in packer.nvim disappears and nvim-tree works correctly after this changes.

### DIFF
--- a/lua/plugins/configs/nvimtree.lua
+++ b/lua/plugins/configs/nvimtree.lua
@@ -84,7 +84,7 @@ local options = {
 }
 
 -- check for any override
-options = require("core.utils").load_override(options, "kyazdani42/nvim-tree.lua")
+options = require("core.utils").load_override(options, "nvim-tree/nvim-tree.lua")
 vim.g.nvimtree_side = options.view.side
 
 nvimtree.setup(options)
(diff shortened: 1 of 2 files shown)


lucario387:
Do the same for nvim-web-devicons

lucario387:
And if possible as well, do another commit to v2.0 branch (just to raise notice in case siduck forgot to cherry-pick)

anosillus:
I see.  I will create two additional pull requests within half a day.

anosillus:
I make #1745.
#1745 fix all kyazdani42 repositories.

PR #1747, opened 2023-01-17
Adds Example Configs to Readme

peterpme:
Getting started can be hard if there's something you're struggling to understand. I went through the top contributors to nvchad and tried to find their configs. Then I added myself.


siduck:
hi @peterpme , check https://github.com/NvChad/example_config

siduck:
I'll include it in the readme from v2.0, just wait till february :D

PR #1770, opened 2023-02-12
Fix sumneko_lua deprecation, change to lua_ls

ariefra:


siduck:
@ariefra can you do it for main branch too?

PR #1771, opened 2023-02-12
fix: sumneko_lua is deprecated

SAJIBxD:


siduck (approved):

georgejean:
The [doc](https://nvchad.com/docs/debugging-config) should be updated too.

PR #1797, opened 2023-03-02
Fix telescope and nvdash search icon, nvdash themes desc

reaink:
nerd font default not view current config search icon
current search icon:   
chagned serach icon:    origin nerdfont icon name nf-seti-search

origin
image
fixed  search icon and open telescope themes desc  <leader>th
image

telescope prefix icon 
image

image

### DIFF
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -60,11 +60,11 @@ M.ui = {
     },
 
     buttons = {
-      { "  Find File", "Spc f f", "Telescope find_files" },
+      { "  Find File", "Spc f f", "Telescope find_files" },
       { "  Recent Files", "Spc f o", "Telescope oldfiles" },
       { "  Find Word", "Spc f w", "Telescope live_grep" },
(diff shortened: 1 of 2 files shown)


siduck:
@reaink are you using iterm?

reaink:
> @reaink are you using iterm?

Yes wezterm can display normally

dat-le-4:
I have the same issue and I'm using iterm2. How to fix this? I'm using hack nerd font (not mono).

siduck:
> I have the same issue and I'm using iterm2. How to fix this? I'm using hack nerd font (not mono).

hi @dat-le-4 either use a different terminal or just override telescope config

dat-le-4:
can you elaborate more on what to override and at which files? That would be very helpful. Thank you.

siduck:
> can you elaborate more on what to override and at which files? That would be very helpful. Thank you.

you have to override telescope config, read https://nvchad.com/#/docs/config/plugins. Then you'd know how to do it

PR #1843, opened 2023-03-19
fix require on line 79 of utils.lua

LowSpecGamerROBLOX:
nv chad was broken out of the box for me so I had to go ahead and fix it up manually. this was done by editing the require statement on line 79 of core/utils.lua to just require the mappings.lua file

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -73,7 +73,7 @@ M.load_mappings = function(section, mapping_opt)
     end
   end
 
-  local mappings = require("core.utils").load_config().mappings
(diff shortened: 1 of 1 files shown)


siduck:
not needed . just do rm -rf .local/share/nvim then reinstall nvchad

LowSpecGamerROBLOX:
ohhhhhh. mind adding that anywhere in the documentation? i couldn't find that anywhere

siduck:
> ohhhhhh. mind adding that anywhere in the documentation? i couldn't find that anywhere

image

PR #1849, opened 2023-03-23
Add NvimTreeFindFileToggle as cmd

dgmora:
[`NvimTreeFindFileToggle`](https://github.com/nvim-tree/nvim-tree.lua/blob/master/doc/nvim-tree-lua.txt#L150) opens the tree with the current file selected, so I like it more than NvimTreeToggle. The issue is that when you open nvim, the command isn't available on startup. cmd isn't configurable, so it'd be great if we could add the command for everyone

### DIFF
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -201,7 +201,7 @@ local default_plugins = {
   -- file managing , picker etc
   {
     "nvim-tree/nvim-tree.lua",
-    cmd = { "NvimTreeToggle", "NvimTreeFocus" },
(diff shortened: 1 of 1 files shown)


siduck:
hi @dgmora , add this in your custom plugins table 

```lua
{
    "nvim-tree/nvim-tree.lua",
    cmd = { "NvimTreeFindFileToggle" },
  },
  ```

dgmora:
I tried something similar with v1 and it didn't seem to work, but with v2 it worked, thanks!

siduck:
> I tried something similar with v1 and it didn't seem to work, but with v2 it worked, thanks!

v2 uses lazy.nvim which merges user plugin table very nicely, so use v2 :D

PR #1890, opened 2023-04-05
fix(mappings): fix description of j key in x mode

prometheusalpha:
Hi, this is my first PR in this repository.
When viewing the default mappings in the config, I noticed that the j key had a description of "move left" while the k key had a description of "move down". So I figured there may have been a mismatch and have changed the description of j key back to "move up".
If my PR format has any problem, please let me know!
Thank you

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -57,7 +57,7 @@ M.general = {
   },
 
   x = {
-    ["j"] = { 'v:count || mode(1)[0:1] == "no" ? "j" : "gj"', "move left", opts = { expr = true } },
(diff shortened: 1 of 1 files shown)


siduck (approved):

siduck:
@prometheusalpha  can you resend the PR? I just realized some mistakes in our mappings file ( before this PR )

image

just replace those, j = down and k = up

and could you add descriptions for all these 4 mappings ( normal mode ) too?

image

siduck:
and could you add descriptions for all these 4 mappings ( normal mode ) too?

image

PR #1902, opened 2023-04-07
Added fuzzy finder for open buffer

sk02241994:
Added a fuzzy finder key mapping for current buffer using Telescope for easy search in current buffer.


lucario387:
Rebase please

PR #1919, opened 2023-04-12
feat(telescope, mappings): add find by selected text

PaulosSouza:


siduck:
hi @PaulosSouza , i think we have enough telescope mappings now. I think this can go in your custom mappings file :)

PR #1967, opened 2023-04-27
Update init.lua : automatically change dir to the file's path.

### LINKED ISSUE #1966: nvim-tree doesn't change dir when other file (path) is opened!
NvChad doesn't update the nvim-tree buffer when a new location is opened in the same terminal path.

**The Expected behavior**
When I cd into another location, the nvim-tree buffer should start from that point.
I have read about `audochdir` but I don't know how to apply!, also a quick fix is to `:cd %:h` from the nvim-tree buffer.

**Screenshots**
nvim-tree_weird

**Desktop:**
 - OS : `Arch Linux BTW`
 - Terminal : `kitty`
 - Nvim : `v0.8.3`

AYehia0:
Change the current working directory whenever you open a file, switch buffers, delete a buffer or open/close a window.  
Fixes: #1966

### DIFF
--- a/lua/core/init.lua
+++ b/lua/core/init.lua
@@ -42,6 +42,9 @@ opt.termguicolors = true
 opt.timeoutlen = 400
 opt.undofile = true
 
+-- automatically change dir
+opt.autochdir = true
+
 -- interval for writing swap file to disk, also used by gitsigns
 opt.updatetime = 250
 


siduck:
hi @AYehia0  i think this is more of user preference! and can just go to your custom/init.lua

Also if I open a file which is in nested subdirs of my main dir ( lava-platform here )

image

then it still cd's into the parent dir. It'd    be good if it checked the dir with .git and then cd into the project!

AYehia0:
I think you're right, but I would expect the normal behavior to be like this : 
working-nvim-tree

I achieved this by not using `autochdir` since it does have some weird behaviors on plugins, instead I created [autocmd](https://vim.fandom.com/wiki/Set_working_directory_to_the_current_file).

```lua

-- autocmd BufEnter * silent! lcd %:p:h
-- the same as autochdir but better for nvim-tree and other plugins.
autocmd("BufEnter", {
    pattern = "*",
    command = "silent! lcd %:p:h"
})

```

AYehia0:
however this works with : `update_root = true` but unfortunately, it goes deep inside nested dirs just like the picture you sent!
I don't know if there is a solution to automatically update the nvim-tree buffer to full path, when switching between tabs ? 
that would be a nice feature to have :D

siduck:
> however this works with : `update_root = true` but unfortunately, it goes deep inside nested dirs just like the picture you sent! I don't know if there is a solution to automatically update the nvim-tree buffer to full path, when switching between tabs ? that would be a nice feature to have :D

sounds good, will add in v3.0

PR #2027, opened 2023-05-16
style(mason): One style icons

LexeyKhom:
(With lazy-nvim icons)


siduck:
hi @LexeyKhom i like our current mason icons, so cant change them. Thanks for the PR tho

Issue #2025, opened 2023-05-15
Error executing vim.schedule lua callback: linux mint 21.1/nvim.0.9


syaoransan:
I just upgraded neovim to version 0.9 from 0.8.3 and when i reinstalled nvchad i got this error

Error executing vim.schedule lua callback: /home/san/.config/nvim/lua/core/utils.lua:80: attempt to index a nil value
stack traceback:
        /home/san/.config/nvim/lua/core/utils.lua:80: in function </home/san/.config/nvim/lua/core/utils.lua:55>

![Screenshot from 2023-05-15 23-35-01](https://github.com/NvChad/NvChad/assets/78855146/cbea5f30-1498-44cc-855f-634090e9d6ae)

siduck:
rm -rf .local/share/nvim

PR #2037, opened 2023-05-19
Fix default mappings not load if chadrc.lua does not exist

### LINKED ISSUE #1961: Error executing vim.schedule lua callback
Error executing vim.schedule lua callback: /home/sabi/.config/nvim/lua/core/utils.lua:80: attempt to index a nil value                                         
stack traceback:                                                                                                                                               
        /home/sabi/.config/nvim/lua/core/utils.lua:80: in function </home/sabi/.config/nvim/lua/core/utils.lua:55>

I had installed NvChad and removed if by  `rm ~/.config/nvim`

Now when I tried to reinstall again, it gives me this error. Whenever I open with NvChad config this error comes

image

azuwis:
The regression was introduced by dab9171e6426b6eb8f0f8bcd0829a0415d260fcb

Fix #1961 #2025

### DIFF
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -87,6 +87,6 @@ M.plugins = "" -- path i.e "custom.plugins", so make custom/plugins.lua file
 
 M.lazy_nvim = require "plugins.configs.lazy_nvim" -- config for lazy.nvim startup options
 
-M.mappings = {}
+M.mappings = require "core.mappings"
 
 return M
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -4,17 +4,15 @@ local merge_tb = vim.tbl_deep_extend
 M.load_config = function()
   local config = require "core.default_config"
   local chadrc_path = vim.api.nvim_get_runtime_file("lua/custom/chadrc.lua", false)[1]
-  local default_mappings = require "core.mappings"
-  config.mappings = default_mappings
 
   if chadrc_path then
     local chadrc = dofile(chadrc_path)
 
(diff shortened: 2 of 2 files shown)


azuwis:
@siduck Can you please have a look at this, `config.mappings` is an empty map if chadrc.lua not exist.

siduck:
> The regression was introduced by [dab9171](https://github.com/NvChad/NvChad/commit/dab9171e6426b6eb8f0f8bcd0829a0415d260fcb)
> 
> Fix #1961 #2025

those errors get fixed when we delete the old nvim dir, it usually happens when you dont delete old nvim dir ( .local/share/nvim ) before installing nvchad

azuwis:
The steps to reproduce is simple:

```
$ rm ~/.config/nvim/lua/custom/chadrc.lua
$ nvim
```

Before dab9171, it's fine because default mapping is loaded in `default_config.lua` (`M.mappings = require "core.mappings"`).

The ability to run NvChad without `chadrc.lua` is necessary for packaging for Linux dist.

I was trying to [package NvChad for NixOS](https://github.com/azuwis/nix-config/blob/master/modules/home/nvchad/default.nix), and found the problem.

siduck:
> The ability to run NvChad without chadrc.lua is necessary for packaging for Linux dist.

you're right. I before removed that code cuz i assumed everyone would have the chadrc by default, because during the installation,  it'll prompt you to bootstrap example_config, if you select no to it, then it'll still create a custom chadrc

siduck (approved):

PR #2062, opened 2023-05-29
Fix Comment.nvim not loading in visual mode

martin-braun:
Adds all required mappings to commenting stuff in and out (incl. from visual mode), to properly lazy load Comments.nvim.

### DIFF
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -200,7 +200,12 @@ local default_plugins = {
 
   {
     "numToStr/Comment.nvim",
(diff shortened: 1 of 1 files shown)


siduck (approved):

willer-rodrigo:
It's missing commas on lines 204 and 208...

martin-braun:
@willer-rodrigo Yes I know, was a bit too quick on that one and should've tested it. Fixed my mistake immediately with #2063, but now @siduck isn't here right now. :/

Sorry about that.

PR #2075, opened 2023-05-31
I want to provide good functionality

josuerom:
I propose that you review the new functionality that you add to the configuration of your beautiful editor, it is to be able to compile and execute the programs written in java, cpp, py, js or ts and more within the same editor, thus making it an IDE, the file is in: lua/core/compiler.lua and import it from the general init.lua


siduck:
@josuerom  huge red flag!! please read https://nvchad.com/docs/contribute .

if you want to share your config then do it in our discord server's share config channel. I'll soon make a webpage on the site where users could share their configs

PR #2083, opened 2023-06-03
Change cmp to insert completion instead of replacing the current word

AndreJesusBrito:
I find this annoying and if we want to actually delete the word we can simply just dw in vim


siduck:
@AndreJesusBrito could you show a video of changes with/without your PR?

AndreJesusBrito:
@siduck sure

btw, what is the tool that you use to record the previews?

siduck:
> @siduck sure
> 
> btw, what is the tool that you use to record the previews?

simplescreenrecorder for quick recording but I use ffmpeg most of the time

AndreJesusBrito:
Before:
before

After:
after

siduck (approved):

PR #2095, opened 2023-06-07
Change `Tab` cmp behaviour

KorigamiK:
The default tab behavior of `nvim-cmp` is to only select not insert the suggestion on arrow movement.
However, since `NvChad` doesn't map `up` and `down` arrow keys `<C-n` and `<C-p>` work the same as `<Tab>` and `<S-Tab>`.

Different keys for the same action should be avoided so, having the `Tab` keys work the same as arrow keys seems like the more acceptable way.

### DIFF
--- a/lua/plugins/configs/cmp.lua
+++ b/lua/plugins/configs/cmp.lua
@@ -81,7 +81,7 @@ local options = {
     },
     ["<Tab>"] = cmp.mapping(function(fallback)
       if cmp.visible() then
-        cmp.select_next_item()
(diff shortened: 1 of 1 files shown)


siduck:
now tab no longer changes the text in insert mode :/

siduck:
i mapped C-n/p before because all menus have C-n/p working by default

KorigamiK:
Hmm, I understand but I prefer having an option of not inserting if I wish so. 
Can you give me an example that I can use in my custom config to achieve this?

Issue #2084, opened 2023-06-04
Telescope syntax highlighting is not working until I open a file


Shunseii:
**Describe the bug**
The syntax highlighting in the Telescope window is not respecting my configuration and uses the default, high-contrast colours when I first open it after opening a project. However, if I open a file and then open Telescope again, the syntax highlighting is back to normal. It's happening with all the languages that I've checked.

**To Reproduce**
1. Navigate to a project directory
2. Open the directory in neovim by running `nvim`
3. Open Telescope with, say `<leader>ff`
4. See the default syntax highlighting 

**Expected behavior**
The syntax highlighting should be consistent and correct in the Telescope window.

**Screenshots**
When I initially open Telescope before opening any project files (shows the broken syntax highlighting)
![image](https://github.com/NvChad/NvChad/assets/20803729/dc67d1a5-ac91-44b7-9e9b-0ae0b05e3b0d)

After I open some file in the project then open Telescope again (shows the expected highlighting):
![image](https://github.com/NvChad/NvChad/assets/20803729/840444ba-8e68-4667-b005-cd94e0686161)

**Desktop (please complete the following information):**
 - Operating System: Pop!_OS 22.04
 - Terminal: Gnome Terminal
 - Neovim Version: V0.9.0

**Additional context**
I have the exact same setup on my laptop (issue occurs on my desktop) in terms of OS and NvChad configuration and I don't see it on my laptop. In fact, I wasn't even seeing this on my desktop until I installed the following LSPs with Mason by adding the following lines to `ensure_installed` and then `:MasonInstallAll` (since I realized I didn't have them installed nor explicitly required in my config):
- "elixir-ls",
- "eslint-lsp",
- "julia-lsp",
- "rust-analyzer",
- "tailwindcss-language-server",
- "typescript-language-server",

I've also copied over the entire `custom/` directory from my laptop to my desktop (although they should've been identical from the start) to double-check, but I'm facing the same issue.

I suspect I just missed a step somewhere when I was setting it up again on my desktop (which I did after my laptop) rather than this being a bug, but I'm really at a loss as to how I can debug this :(

siduck:
unable to reproduce! 

make sure you have these tsparsers installed :

html css javascript  typescript tsx

Shunseii:
I verified just now with `:TSInstallInfo` that I have them all installed. The syntax highlighting works in almost all cases, it's just when I open the Telescope window before I open any file -- it's as if something doesn't load until I open a file? Really strange

[...]

PR #2097, opened 2023-06-08
telescope: Explicitly depend on nvim-treesitter

azuwis:
Fix Telescope syntax highlighting is not working until open a file https://github.com/NvChad/NvChad/issues/2084


siduck:
i'll have to wait! i cant seem to reproduce the issue at all https://github.com/NvChad/NvChad/issues/2084#issuecomment-1579793113

Issue #1868, opened 2023-03-28
Feature: Provide more overrides for configuring cmp


airtonix:
**Is your feature request related to a problem? Please describe.**

1. currently the cmp plugin can only be affected from the `chadrc` with regards to theming: 
    -  https://github.com/NvChad/NvChad/blob/v2.0/lua/plugins/init.lua#L186-L188C9
    - https://github.com/NvChad/NvChad/blob/v2.0/lua/plugins/configs/cmp.lua#L5
2. vbut there's a bunch of things users would like to customise here that aren't affected by the chardrc: 
    - https://github.com/NvChad/NvChad/blob/v2.0/lua/plugins/configs/cmp.lua#L70-L113 

**Describe the solution you'd like**

some way to provide more overrides to the the rest of cmp via `chadrc`

**Describe alternatives you've considered**

- copy paste the cmp plugin entry along with all related files into `custom`

**Additional context**

i want to change the order of the completions so that lsp is first.

**Screenshot**
Maybe a screenshot of the feature

siduck:
@airtonix those things can be overriden by the "opts" field!

![image](https://user-images.githubusercontent.com/59060246/228384564-7529948e-54e2-481b-9479-acd8ead0118b.png)

airtonix:
@siduck But wont doing it exactly like that blow away all the cmp plugin setup that already outside of the custom directory?

airtonix:
nvm ok looks like it might be merging it.

airtonix:
Ok so the reason why i think i opened this ticket was this comment in the docs: 

https://github.com/NvChad/nvchad.github.io/blob/src/src/routes/(index)/docs/config/plugins.mdx?plain=1#L43


![image](https://user-images.githubusercontent.com/61225/228677127-f601397a-d2a0-44ce-9067-f9aa155e8c2e.png)


But it looks like you've just suggested to me somethign that contradicts that ?

siduck:
> @siduck But wont doing it exactly like that blow away all the cmp plugin setup that already outside of the custom directory?

it wont, lazy.nvim overrides it.

siduck:
> Ok so the reason why i think i opened this ticket was this comment in the docs:
> 
> https://github.com/NvChad/nvchad.github.io/blob/src/src/routes/(index)/docs/config/plugins.mdx?plain=1#L43
> 
[...]

PR #2117, opened 2023-06-14
feat: update cmp.lua to remove tab binds

aneshodza:
Tab and Shift tab being bound to swapping trough the suggestions is pretty annoying, so maybe we change it to be voluntary instead of default?
I had to do [this](https://github.com/aneshodza/nvchad_config/blob/eaf1f41626afcec03ac91aad1a45444c0e0ace9d/configs/overrides.lua#L77) to fix it.


siduck:
only    you find it annoying, so disable it at your end https://github.com/NvChad/NvChad/issues/1868

PR #2158, opened 2023-06-27
Add mapping for LSP formatting command

limitedink:
The specific change made:
- Added a mapping for the LSP formatting command ("<leader>fm") to invoke the "vim.lsp.buf.format()" function.
- The default mapping was not working, so this custom mapping was added to ensure the functionality is available.

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -32,7 +32,8 @@ M.general = {
     -- line numbers
     ["<leader>n"] = { "<cmd> set nu! <CR>", "Toggle line number" },
     ["<leader>rn"] = { "<cmd> set rnu! <CR>", "Toggle relative number" },
-
+    --LSP formatting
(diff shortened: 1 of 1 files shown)


georgejean:
There is already this mapping in `core/mappings.lua`
```lua
    ["<leader>fm"] = {
      function()
        vim.lsp.buf.format { async = true }
      end,
      "LSP formatting",
    },
```
but there are also these two lines in `plugins/configs/lspconfig.lua`:
```lua
client.server_capabilities.documentFormattingProvider = false
client.server_capabilities.documentRangeFormattingProvider = false
```

If you want to enable LSP formatting, you can use in your `custom/configs/lspconfig.lua`:

```lua
local on_attach_bis = function(client, bufnr)
  require("plugins.configs.lspconfig").on_attach(client, bufnr)
  client.server_capabilities.documentFormattingProvider = true
  -- client.server_capabilities.documentRangeFormattingProvider = true ( if possible )
end

lspconfig["your_lsp_client_name"].setup {
  on_attach = on_attach_bis,
  ...
  }
```

PR #2162, opened 2023-06-27
Pull Request: Allow Customization of "mason/bin" Path Handling

TBytes404:
### Description
This pull request introduces a new feature that allows users to customize the handling of the "mason/bin" path in the system. The change is achieved by adding the `place_mason_bin` option to the [`custom/chadrc.lua` file][1]. Users can now choose whether to prepend or append the "mason/bin" path based on their specific needs.

I have set the default option to prepend, aligning with the existing setting used in [`mason.nvim`][2] for consistency.

This change was prompted by a [discussion][3] regarding the need for more flexibility in managing the "mason/bin" path.

### Request
I kindly request a thorough review and verification of this pull request. Your feedback and suggestions are highly appreciated. Thank you!

[1]:https://github.com/TirtharajPramanik/example_config/blob/d19060f029642f039604d2de12b91270c0e52720/chadrc.lua#L15C1-L21
[2]:https://github.com/williamboman/mason.nvim/blob/db58ec7bc9248fb1878172f90cadc825b77564a6/lua/mason/settings.lua#L10-L15
[3]:https://github.com/NvChad/NvChad/pull/2031#issuecomment-1606449443

### DIFF
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -83,12 +83,9 @@ M.ui = {
   },
 }
 
--- Where NvChad should put mason.nvim bin location in your PATH. Can be one of:
--- - "prepend" (default, Mason's bin location is put first in PATH)
--- - "append" (Mason's bin location is put at the end of PATH)
--- - "skip" (doesn't modify PATH)
+-- Determines the preferred placement of the mason.nvim bin directory in your PATH.
 ---@type '"prepend"' | '"append"' | '"skip"'
-M.mason_path = "prepend"
+M.place_mason_bin = "prepend"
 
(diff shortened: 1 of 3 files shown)


adamrp:
This looks useful, thanks @TirtharajPramanik! Recommend also adding comments in `lua/plugins/config/mason.lua` as well as `lua/plugins/init.lua` indicating that the user should not attempt to set the `PATH` option via the plugin's settings since doing so will not play nicely with the explicit path manipulation in `lua/core/init.lua`

TBytes404 (commented):
Enhanced code with descriptive comments to alert users about potential tradeoffs when customizing options

PR #2163, opened 2023-06-27
Use mason config option for PATH

adamrp:
This PR removes the PATH manipulation in `core/init.lua` that puts the `mason/bin` path onto users' `PATH` which can mask or even override the user's specified config in `custom/plugins.lua`.

Instead, the plugin's builtin `PATH` option is leveraged, which is in better alignment with how every other plugin is configured (default config is overridden by/merged wth user `custom` config).

To accomplish this, this PR also loads mason eagerly (i.e. `lazy=false`) so that the path is always updated in the manner requested by the user.


siduck:
i dont want to remove lazyloading of mason plugin, thats the reason i added its cmds and path

PR #2215, opened 2023-07-16
fix(icon): replace find icon with f021e

awe123343:
The previous find fine icon does not work in iTerm2 even if using nerd font. Replacing it with f021e
After fix:
image
Before fix:
image

### DIFF
--- a/lua/core/default_config.lua
+++ b/lua/core/default_config.lua
@@ -63,7 +63,7 @@ M.ui = {
     },
 
     buttons = {
-      { "  Find File", "Spc f f", "Telescope find_files" },
+      { "󰈞  Find File", "Spc f f", "Telescope find_files" },
       { "󰈚  Recent Files", "Spc f o", "Telescope oldfiles" },
       { "󰈭  Find Word", "Spc f w", "Telescope live_grep" },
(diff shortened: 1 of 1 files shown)


siduck:
@awe123343 thats an iterm issue! you can just replace that in your config. these icons look better so I use it

image

antoineco:
@awe123343 there were a lot of changes between Nerd Font v2 and v3 (released a few weeks ago). Can you confirm you're using a recent version of "Nerd Font Symbols"?

awe123343:
> @awe123343 there were a lot of changes between Nerd Font v2 and v3 (released a few weeks ago). Can you confirm you're using a recent version of "Nerd Font Symbols"?

Yeah, I am using v3

awe123343:
> @awe123343 thats an iterm issue! you can just replace that in your config. these icons look better so I use it
> 
> image

gotcha, thx for explaining. yeah, I already replaced in my config

siduck:
> > @awe123343 thats an iterm issue! you can just replace that in your config. these icons look better so I use it
> > image
> 
> gotcha, thx for explaining. yeah, I already replaced in my config

or you could use terminals like alacritty, wezterm, kitty. Ive heard a lot of hatred against iterm2 before lol

PR #2271, opened 2023-08-07
feat: added .ignore file

invisible1988:
Added .ignore file. So telescope doesn't ignore the custom folder.

### DIFF
--- a/.ignore
+++ b/.ignore
@@ -1,0 +1,1 @@
+!/lua/custom/


georgejean:
It would be nice to add your .ignore in .gitignore too (for possible additions of files/folders).

georgejean:
Another option (I don't know if it works) would be to add .ignore in .gitignore to allow further possible file/folder additions.

invisible1988:
Hi @georgejean,

do you mean removing the `.ignore` file and adding an exclusion for `.ignore` to `.gitignore`? So that any user could do it themselves? Or do you mean to keep the `.ignore` file and add it to `.gitignore`? So it would be shipped with the branch, but it would be possible to customise it?

Both would be possible....

georgejean:
Hi @invisible1988.
I thought about the first option but if the second works, I think it's better (because of the included example).
I let @siduck give his opinion.

invisible1988:
I think to ship a `.ignore` file would be good. But let's see what @siduck thinks. I'm fine with both.

georgejean:
In fact I think the second option
> Or do you mean to keep the `.ignore` file and add it to `.gitignore`? So it would be shipped with the branch, but it would be possible to customise it?

wouldn't have downloaded the `.ignore` file with a `nvchadupdate` (because `.ignore` would be gitignored). Only fresh nvchad users would have this file (`git clone ...`). It's better now but it would be interesting to be able to add some of the other folder in `.gitignore` though.

PR #2310, opened 2023-08-19
fix: fixed invalid index referencing in table due to user-specified i…

bozkayasalihx:
* its fixed "attempt to index a nil value" when used passed invalid configurations in his/her `custom/` directory


siduck:
whats this? 
image

bozkayasalihx:
@siduck, i assume i've accidentally created that file, I will remove it right now

siduck:
and how do you reproduce the bug?

bozkayasalihx:
@siduck, sure i was working on a golang project.  i did want to install dap for golang debuger but i didn't want to configure custom mapping for that just want to go with default configuration. after install dap plugin chadvim immediately threw `attempt to index nil value` in "core.utils.lua", i'll attach to ss for in case

Screenshot 2023-08-21 at 12 39 08
 
Screenshot 2023-08-21 at 12 44 07

but whenever i configured to dap plugin custom mapping. that error went away. here is my config
btw, this config is in `custom/mappings.lua`

Screenshot 2023-08-21 at 12 45 42

siduck:
@bozkayasalihx you should format files using our stylua config btw

bozkayasalihx:
@siduck, I thought i did. So yeah i've reformatted the files with stylua and committed too

siduck:
can you share your mapping which causes this bug?

>  i didn't want to configure custom mapping for that just want to go with default configuration

btw you're supposed to add all your changes in custom dir only

bozkayasalihx:
I already added, look at the screenshots and BTW I know where to put all of custom configuration but thanks for reminding

siduck:
image

I added these mappings, i dont face any errors?

Issue #1811, opened 2023-03-12
Help with custom mappings - different mapping based on filetype


MineRock:
Hello, I really liked this project and am using it as my default config right now. I wanted to make a Compile-and-Run key mapping so it becomes easier for me to code and I made this to detect filetype in custom/mappings.lua
```
local function filetype()
  if vim.cmd("echo &filetype") == "rust"
  then
    return {"<cmd> term rustc % && %:r <CR>", "run rust program"}
  else
    return {"<cmd> lua echo vim.bo.filetype<CR>", vim.cmd("echo &filetype")}
  end
end
```
In M.general, I've set the mapping as follows, 
```
["<leader><F10>"] = filetype(),
```
But the mapping doesn't work. I changed the else command to the above to see what it was returning and it just returned a blank value. I perhaps need the mapping to load after the filetype has loaded in, but I have no clue how I could do that. Could you help me or direct me to some resources? Thanks!

Edit: I am on v2.0 branch right now.

ayush-india:
Try to follow the mapping guide lines (how to make custom mappings)

siduck:
```lua

M.blabla = {

  n = {

    ["<leader>cc"] = {
      function()
        if vim.bo.ft == "rust" then
          vim.cmd "term rustc % && %:r"
        else
          vim.notify(vim.bo.ft)
        end
      end,

      vim.bo.ft == "rust" and "run rust program" or vim.bo.ft,
    },
  },
}
```

igordertigor:
I have a similar issue, but @siduck's solution doesn't quite work. Unfortunately, the custom mappings guide is relatively silent about this as well. I have a bunch of shortcuts for unicode characters, that I want to activate when typing markdown, but not elsewhere (ideally, I would even want them turned off when I'm in a markdown code block). As a consequence, I can't call a function when hitting the respective key, but I just don't want the key to be mapped at all. I would expect that this would somehow need to modifiy the `M` table, but I can't quite figure out how. Here is what I have:
```lua
M.markdown = {
  i = {
    ["->"] = {"→", "right arrow", opts = {nowait = true}},
    ["~>"] = {"⇝", "leads to arrow", opts = {nowait = true}},
    ["=>"] = {"⇒", "implies arrow", opts = {nowait = true}},
  }
}
```
Shortened a bit, but you get the point. The mappings work, but they work for all files.

[...]

PR #2335, opened 2023-08-25
Remove mappings feature

BrunoKrugel:
Created this function so users can load/unload mappings on demand, this can be usefull when creating mappings for specific filetypes.

For example if you want a keybind to be available only in `go` filetype, you create your mapping with the `plugin` flag

```
M.go = {
  plugin = true,
  n = {
    ["<F1>"] = { "do something", "Text" },
  },
}
```

And to load/unload the keybind you can create an autocmd that will load and unload the mappings:

```
local autocmd = vim.api.nvim_create_autocmd
autocmd("FileType", {
  callback = function()
    if vim.bo.ft == "go" then
      require("core.utils").load_mappings "go"
    else
      require("core.utils").remove_mappings "go"
    end
  end,
})
```

In this way, we keep the record of the mappings in nvcheatsheet, maybe we can create a function to update nvcheatsheet if the mapping is loaded/unloaded?

I am creating this PR because I thought this could be usefull for others users  #1811, but I dont know if this can be improved or if there is another way to do it.

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -87,6 +87,33 @@ M.load_mappings = function(section, mapping_opt)
   end)
 end
 
+M.remove_mappings = function(section)
+  vim.schedule(function()
+    local function remove_section_map(section_values)
+      if section_values.plugin then
+        return
+      end
+
+      for mode, mode_values in pairs(section_values) do
+        for keybind, _ in pairs(mode_values) do
+          local _, _ = pcall(vim.api.nvim_del_keymap, mode, keybind)
+        end
+      end
+    end
+
+    local mappings = require("core.utils").load_config().mappings
+
+    if type(section) == "string" then
+      mappings[section]["plugin"] = nil
+      mappings = { mappings[section] }
+    end
+
+    for _, sect in pairs(mappings) do
+      remove_section_map(sect)
+    end
+  end)
+end
+
 M.lazy_load = function(plugin)
   vim.api.nvim_create_autocmd({ "BufRead", "BufWinEnter", "BufNewFile" }, {
     group = vim.api.nvim_create_augroup("BeLazyOnFileOpen" .. plugin, {}),


georgejean:
Hi Bruno,
> but I dont know if this can be improved or if **there is another way to do it**.

Another way to use keymappings for specific filetypes is to use the `ftplugin` folder (see [there](https://www.reddit.com/r/neovim/comments/x3zp6t/usage_of_afterftplugin_directory_for/)). I use it with NvChad directly in `~/.config/nvim` folder for `.tex` files because `ftplugin` folder is gitignored here (I could have used `after/ftplugin` too), so I have :
image
The `tex.vim` file contains my keybinds for `.tex` files (It could also be a `tex.lua` file).
It works for my simple needs (but the keymappings aren't registered in the cheatsheet).

igordertigor:
I like @BrunoKrugel 's approach, because it keeps the mapping definitions together, but I agree with @georgejean , that the ftplugin approach would be more standard vim/nvim.

siduck:
i think this can be just added to our api docs 

https://nvchad.com/docs/api

BrunoKrugel:
> i think this can be just added to our api docs 
> 
> 
> 
> https://nvchad.com/docs/api

That was my intention, after the merge, edit the API documentation

BrunoKrugel:
> I like @BrunoKrugel 's approach, because it keeps the mapping definitions together, but I agree with @georgejean , that the ftplugin approach would be more standard vim/nvim.

Yes, his approach is more easier and more standard, but the mappings won't appear in nvcheatsheet.

Also, if the mappings are loaded in this approach, if we open another file (another file type) the mappings will still be loaded, right?

siduck:
> > i think this can be just added to our api docs
> > https://nvchad.com/docs/api
> 
> That was my intention, after the merge, edit the API documentation

its not that the API docs only has stuffs that nvchad's functions provide, you can define new functions in the api doc ( as long as they're related to nvchad behavior ) , so i dont think we need to add code for this in nvchad repo.

PR #2348, opened 2023-08-30
add "<leader>pp" to popup "List projects" window.

trujunzhang:
add "<leader>pp" to popup "List projects" window.

I was a Lunarvin user before, and lunarvim set up project management as default, 
and now I use nvchad, I want Nvchad like Lunarvim to have project management,  and other Nvchad's users also think so like me.
So I suggest adding project management in the Nvchad's core project.


siduck:
https://nvchad.com/docs/contribute

PR #2353, opened 2023-09-02
feat(lsp): add visual mode code actions

orhnk:
### V-mode code actions

Some LSP servers come with refactoring tools out of the box.
Actions like `move to function` only works in visual mode (e.g rust-analyzer)

I think this feature is generic enough to get added as a visual-mode keymapping
> [!NOTE]
> It's all built-in lsp server

### DEMO

https://github.com/NvChad/NvChad/assets/101834410/c0036e6e-b377-41bb-9b52-a7d122caf075

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -240,6 +240,15 @@ M.lspconfig = {
       "List workspace folders",
     },
   },
+
+  v = {
+    ["<leader>ca"] = {
+      function()
+        vim.lsp.buf.code_action()
+      end,
+      "LSP code action",
+    },
+  },
 }
 
 M.nvimtree = {


siduck:
> Actions like move to function only works in visual mode (e.g rust-analyzer)

this seems like language specific to me, if all other lsps dont support this thing then i dont think that mapping's useful. You can just add it in your custom mappings

orhnk:
This is a wanted feature in a wide range of languages.

I didn't check them from their docs but here are some I've just tried:


> [!NOTE]
> Some of these are coming from a 3rd party plugin: `refactoring.nvim`
> (If you see null-ls after the action)

### Python

image

### Rust

image

### TypeScript

image

### C/C++

image

### Lua

image

### Java

https://github.com/mfussenegger/nvim-jdtls#extensions

And so on...

PR #2360, opened 2023-09-04
performance: avoid reloading config

georgejean:
Each times `require("core.utils").load_config` is called, custom chadrc and custom mappings are merged with the default.
In particular, each call triggers the function `remove_disabled_keys` which merges the mappings.
I think it's not always necessary (excepting in the autocommand in `core.init.lua` which reloads some chadrc options on save), but maybe I'm wrong.
I propose these changes to solve this problem (use config as a property, not a method). The `config` property could perhaps be used in the `ui` repo too.
Below is a test which shows the number of calls before and after the changes.
## Before :
nvchad_load_config1

## After :
nvchad_load_config2

_Note_: I've also added a test to avoid the case where the chadrc file was created but is empty (nil).

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -88,8 +88,7 @@ A fuzzy file finder, picker, sorter, previewer and much more:
 
 - Many beautiful themes, theme toggler by our [base46 plugin](https://github.com/NvChad/base46)
 - Inbuilt terminal toggling & management with [Nvterm](https://github.com/NvChad/nvterm)
-- NvChad updater, hide & unhide terminal buffers with [NvChad extensions](https://github.com/NvChad/extensions)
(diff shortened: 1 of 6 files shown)


siduck:
@georgejean  you should change it at many other places too i.e our plugins (ui & base46)

image

georgejean:
Ok, I will change it and do some testing before making PR at these repos.

siduck:
> Ok, I will change it and do some testing before making PR at these repos.

no worries, i'll do it tomorrow!

siduck:
ig its better to add this feature from v3.0 :D

lucario387:
This makes it so the au in `core/init.lua` doesn't work anymore :thinking: 

I won't fix it :_p:

siduck:
> This makes it so the au in `core/init.lua` doesn't work anymore 🤔
> 
> I won't fix it :_p:

@lucario387 its just in testing so chill, I will not merge it into main unless it works 100% at all places

PR #2371, opened 2023-09-08
add fzf telescope extension to improve sorting performance

kayuxx:
This is a suggestion that was already made in the Discord suggestion section. The addition of [telescope-fzf-native](https://github.com/nvim-telescope/telescope-fzf-native.nvim) can improve performance and sorting behavior (as also recommended by Telescope itself) so, it would be better to include this in **V3.0**.

_**Note**: this extension requires `make` and `gcc`_

### DIFF
--- a/lua/plugins/configs/telescope.lua
+++ b/lua/plugins/configs/telescope.lua
@@ -49,7 +49,7 @@ local options = {
     },
   },
 
-  extensions_list = { "themes", "terms" },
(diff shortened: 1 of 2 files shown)


siduck:
you should set it in extensions too 

image

kayuxx:
@siduck, I just realized that I forgot it.

siduck (approved):

PR #2382, opened 2023-09-12
Keep Visual Selection when indent

a-b:
Features Added:
- When user indent visually selected text with `>>` or `<<` this change will keep visual selection

Reasoning:
Text indentation is used quite often. Many public vim configs contain similar mapping. My understanding is that this was not implemented earlier because `gv` was remapped until recent time.

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -61,6 +61,8 @@ M.general = {
   v = {
     ["<Up>"] = { 'v:count || mode(1)[0:1] == "no" ? "k" : "gk"', "Move up", opts = { expr = true } },
     ["<Down>"] = { 'v:count || mode(1)[0:1] == "no" ? "j" : "gj"', "Move down", opts = { expr = true } },
+    ["<"] = { "<gv", "Indent line" },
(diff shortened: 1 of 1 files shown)


siduck:
could you show behavior with / without this?

a-b:
### Previous behavior
1. Visually select a block of text
2. Press `>` and `>` to indent it to the right
3. Visual selection drops

### Proposed behavior
1. Visually select a block of text
2. Press `>` and `>` to indent it to the right
3. Visual selection stays so that we can repeat step 2 without selecting text again or restore visual selection `gv`

Same works for indentation to the left `<` and `<`

### Links
- https://vimhelp.org/change.txt.html#%3E%3E
- https://vimhelp.org/change.txt.html#%3C%3C
- https://vimhelp.org/visual.txt.html#gv

Please let me know if you have any questions.

siduck (approved):

yusufshalaby:
Just press `.` to repeat the indent

a-b:
@yusufshalaby - you're not wrong; the downside of your solution is that visual selection still goes away -- for huge files that make a difference.

PR #2391, opened 2023-09-16
fix: Improve `remove_disabled_keys` to actually remove keys

kjughx:
There was no way to properly disable a keybinding before. Remapping a keybinding to `{}` only bound it to nothing.
E.g., I have bound  "<leader>c" to close a tab but by default "\<leader\>ch" is bound to NvCheatsheet , so if I press \<leader\>, wait for some time and then 'c', it is interpreted as a prefix. I'd like to unbind "\<leader\>ch" (and all bindings that have "<leader>c" as a prefix) so the above behaviour doesn't happen.

So `remove_disabled_keys()` now actually _removes_ the keys by creating a new table that merges default_mappings and chadrc_mappings but ignores keybindings that are bound to `{}`.

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -1,6 +1,16 @@
 local M = {}
 local merge_tb = vim.tbl_deep_extend
 
+local key_exists = function(mappings, section, mode, key)
+  if not mappings[section] then
+    return false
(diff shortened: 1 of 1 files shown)


siduck:
M.disabled

PR #2443, opened 2023-10-15
Fix unreachable mappings for the lsp

KorigamiK:
since `<leader>ff` exists `<leader>f` cannot happen. so I change it to `<leader>lf`


siduck:
@KorigamiK leader f works here

KorigamiK:
> @KorigamiK leader f works here

Yes, however it's not ideal. There is a delay of at least `1s` before the floating diagnostic window pops up. 
This time is significant enough to be noticeably sluggish. So another keybind is kinda ideal.

PR #2456, opened 2023-10-22
feat(mappings): make <esc> a cmd to remove flickering

TheFedaikin:
This is a small PR to address `Esc` flickering that will occur if you install `dressing.nvim` or any other plugin that will modify the appearance of command-line mode. Unless I missed something functionality is exactly the same.

If needed I can add screenshot, but this should be self-explanatory.

Edit - Screenshot is below:
flicker

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -16,7 +16,7 @@ M.general = {
   },
 
   n = {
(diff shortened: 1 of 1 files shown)


siduck (approved):

BrunoKrugel:
You could achieve this with `opts = { silent = true }`

PR #2490, opened 2023-11-13
remove Search, IncSearch and CurSearch highlights from cmp windows

wochap:
In this commit, the search highlights have been removed from cmp windows to enhance the overall user experience.

Before changes:
Screenshot_2023-11-12_10-32-31

After changes:
Screenshot_2023-11-12_10-30-19

### DIFF
--- a/lua/plugins/configs/cmp.lua
+++ b/lua/plugins/configs/cmp.lua
@@ -52,7 +52,7 @@ local options = {
   window = {
     completion = {
       side_padding = (cmp_style ~= "atom" and cmp_style ~= "atom_colored") and 1 or 0,
-      winhighlight = "Normal:CmpPmenu,CursorLine:CmpSel,Search:PmenuSel,Search:None,IncSearch:None,CurSearch:None",
+      winhighlight = "Normal:CmpPmenu,CursorLine:CmpSel,Search:None,IncSearch:None,CurSearch:None",
       scrollbar = false,
     },
     documentation = {
--- a/lua/plugins/configs/cmp.lua
+++ b/lua/plugins/configs/cmp.lua
@@ -52,12 +52,12 @@ local options = {
   window = {
     completion = {
       side_padding = (cmp_style ~= "atom" and cmp_style ~= "atom_colored") and 1 or 0,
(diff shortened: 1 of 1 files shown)


wochap:
I need to do

siduck:
looks good, but how are u able to search when cmp's visible? it would require exiting insert mode and then using command mod right, by that cmp will auto-close itself!

wochap:
You are correct; however, it is possible to search and have the CMP menu open if you have enabled [completions for the / search based on the current buffer](https://github.com/hrsh7th/cmp-cmdline).

```lua
  cmp.setup.cmdline({ "/", "?" }, {
    mapping = cmdlineMapping,
    sources = {
      { name = "buffer", max_item_count = 10 },
      { name = "cmdline_history", max_item_count = 10 },
    },
  })
```

Before:
Screenshot_2023-11-12_11-07-14

After:
Screenshot_2023-11-12_11-06-43

siduck:
what if we remove the search hl itself 
image

wochap:
That also works; in fact, cmp does that. It only links 'Search' to 'None'. [Source](https://github.com/hrsh7th/nvim-cmp/blob/main/lua/cmp/config/default.lua#L106)

wochap:
I made changes, so only Search is removed, for cmp completion window

siduck:
no i mean removing whole `Search:None` itself,  does cmp link Search to anything by default? if yes then your change is valid

wochap:
> does cmp link Search to anything by default?

through the cmp configuration, yes, the default value for `window.completion.winhighlight` has `Search:None` which remove the Search highlight

```
window.completion.winhighlight = 'Normal:Pmenu,FloatBorder:Pmenu,CursorLine:PmenuSel,Search:None',
```

but nvchad cmp configuration overwrites it, and links Search to PmenuSel:

```
window.completion.winhighlight = "Normal:CmpPmenu,CursorLine:CmpSel,Search:PmenuSel",
```

with `Search:None` (default cmp)
Screenshot_2023-11-12_11-06-43

with `Search:PmenuSel` (nvchad cmp config):
Screenshot_2023-11-12_11-07-14

removing `Search:None` and `Search:PmenuSel`:
Screenshot_2023-11-12_11-10-38

siduck (approved):

PR #2498, opened 2023-11-16
Consistent border style with Mason window

michaelvanstraten:
Quick one—I really love the project, great stuff.

Just noticed an inconsistency that's been bothering me for a while, so I went ahead and fixed it. The border styles are all well and consistent, except when you pop open a Mason window—kind of breaks the flow.

So, my PR's all about making those Mason window borders round like the rest of the gang. Just a small tweak for a smoother look. Take a peek when you get a chance!

### DIFF
--- a/lua/plugins/configs/mason.lua
+++ b/lua/plugins/configs/mason.lua
@@ -4,6 +4,8 @@ local options = {
   PATH = "skip",
 
   ui = {
+    border = "rounded",
+
     icons = {
       package_pending = " ",
       package_installed = "󰄳 ",


michaelvanstraten:
Or should we put this behind a config option like [`telescope.style`](https://github.com/NvChad/NvChad/blob/9d37797e6f9856ef25cfa266cff43f764e828827/lua/core/default_config.lua#L30)?

siduck:
this doesnt require nvchad's base46 stuff , so not needed in chadrc. Also you can just override it with lazy opts

```lua
{ 
 "Mason blabla",
 opts = { here! }
 }
 ```
 
 and put that in custom plugins

PR #2500, opened 2023-11-17
fix default nvtree icon break

phamngocquy:
- the `default` icon in nvimtree.lua is break.
I have try with nerd front but it still break. So, the idea is replace it be using `empty` folder icon
image


phamngocquy:
before:
image



after:
image

siduck:
update your nerdfonts

PR #2528, opened 2023-11-27
Add support for non-interactive bootstrapping

x0wllaar:
Currently, bootstrapping NvChad requires interactive input on the first run, which makes it impossible to use in Dockerfiles. This PR adds a way to answer the "Do you want to install example custom config?" prompt non-interactively via an environment variable.

To run bootstrapping non-interactively, set the `NVCHAD_INSTALL_EXAMPLE_CONFIG` environment variable to the answer you want, and then run nvim, e.g.:
```
NVCHAD_INSTALL_EXAMPLE_CONFIG=no nvim -c "autocmd User PackerComplete quitall" -c "quitall"
```
Which will automatically do all bootstrapping without installing the example config and then immediately exit.

### DIFF
--- a/lua/core/bootstrap.lua
+++ b/lua/core/bootstrap.lua
@@ -1,5 +1,6 @@
 local M = {}
 local fn = vim.fn
+local env = vim.env
 
 M.echo = function(str)
   vim.cmd "redraw"
@@ -9,6 +10,18 @@ end
 local function shell_call(args)
   local output = fn.system(args)
   assert(vim.v.shell_error == 0, "External call failed with error code: " .. vim.v.shell_error .. "\n" .. output)
+end
+
+local function install_example_config_yn()
+  local input = ""
+  if env.NVCHAD_INSTALL_EXAMPLE_CONFIG ~= nil then
+    input = env.NVCHAD_INSTALL_EXAMPLE_CONFIG
+  else
+    input = fn.input "Do you want to install example custom config? (y/N): "
+  end
+  local lower_input = input:lower()
+  local is_yes = lower_input:sub(1,1) == "y"
+  return is_yes
 end
 
 M.lazy = function(install_path)
@@ -40,9 +53,7 @@ M.gen_chadrc_template = function()
   local path = fn.stdpath "config" .. "/lua/custom"
 
   if fn.isdirectory(path) ~= 1 then
-    local input = fn.input "Do you want to install example custom config? (y/N): "
-
-    if input:lower() == "y" then
+    if install_example_config_yn() then
       M.echo "Cloning example custom config repo..."
       shell_call { "git", "clone", "--depth", "1", "https://github.com/NvChad/example_config", path }
       fn.delete(path .. "/.git", "rf")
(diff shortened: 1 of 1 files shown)


siduck:
there should be a cleaner way to do this imo, i'll look into it this weekend

comfysage:
I suggest just changing line 43 to:
```lua
local input = env.NVCHAD_INSTALL_EXAMPLE_CONFIG or fn.input "Do you want to install example custom config? (y/N): "
```
instead of creating an unnecessary function that won't ever be used again

geekobiloba:
This is interesting.

Any update on this workaround?

siduck:
> I suggest just changing line 43 to:
> 
> ```lua
> local input = env.NVCHAD_INSTALL_EXAMPLE_CONFIG or fn.input "Do you want to install example custom config? (y/N): "
> ```
> 
> instead of creating an unnecessary function that won't ever be used again

env is always empty here, how does one set env var?

pwnwriter:
Well, Technically we can use something like 
`nvim  +"NoInput"` or something like that while `NoInput` being a `bool` to check while first installation?

pwnwriter:
Hey, there's a method you can use: `nvim_create_user_command()`. If users provides the cmd, proceed; otherwise, prompt them for input. For example.

```lua
vim.api.nvim_create_user_command("NoInput", function ()
local file = io.open("foo.txt", "w")
if file then
    file:close()
else
    print("Error creating the file.")
end
end,{})
```

This creates a file if they provide `nvim +"NoInput"`

comfysage:
> > I suggest just changing line 43 to:
> > ```lua
> > local input = env.NVCHAD_INSTALL_EXAMPLE_CONFIG or fn.input "Do you want to install example custom config? (y/N): "
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead of creating an unnecessary function that won't ever be used again
> 
> env is always empty here, how does one set env var?

@siduck in this specific codeblock `env` is set to `vim.env` which assumes all environment variables set at runtime.

siduck:
> > > I suggest just changing line 43 to:
> > > ```lua
> > > local input = env.NVCHAD_INSTALL_EXAMPLE_CONFIG or fn.input "Do you want to install example custom config? (y/N): "
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > instead of creating an unnecessary function that won't ever be used again
> > 
> > 
> > env is always empty here, how does one set env var?
> 
> @siduck in this specific codeblock `env` is set to `vim.env` which assumes all environment variables set at runtime.

oh cool, ive to actually read the PR now

siduck (approved):

lucario387:
why again?

`vim.env.NVCHAD_EXAMPLE_CONFIG` can be `yes` and it will fail, even though the intention is probably to still install example config

siduck:
wym again? 
> vim.env.NVCHAD_EXAMPLE_CONFIG can be yes and it will fail

such users are supposed to keep it y or n ,if it fails then its their problem. We have provided env var check then they must follow it

lucario387:
And now the next question. Where is the documentation for the env var?

PR #2538, opened 2023-12-03
[feat] Make gitsigns init function async

### LINKED ISSUE #215: Delay  opening buffers (about 500ms)
Opening an existing file adds about 500ms when NVChad is installed (no extra plugins, as far as I can tell).
This seems to happen when opening buffers: `563.087  545.325: opening buffers`. See attached file at end for complete timing log.

**To Reproduce**
1. Run `nvim somefile.txt`.
2. NVim opens quickly, but there's a noticeable delay before the file appears

**Expected behavior**
Opening files should be real quick

**Desktop (please complete the following information):**
 - Ubuntu 22.10
 - Fish
 - NVChad rev: 9d37797e6f9856ef25cfa266cff43f764e828827 (v2.0 branch)
 - Version of Neovim:

 ```
NVIM v0.10.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

**Additional context**
- I have benchmarked the startup using `--startuptime` flag: 
[timing.txt](https://github.com/NvChad/NvChad/files/13537758/timing.txt)

- File type does not seem to make any difference. I used a txt file, but opening a cpp or a ts file leads to similar results.
- File size doesn't seem to matter either. The test file was 11k in size, but the same happened when I opened a 17 bytes json
- File opens instantly without NVChad
- The test was made on a file that is not inside a Git repo
- Repeating the test leads to similar numbers as well, so it's unlikely to be related to IO waits when reading the file

### PR DESCRIPTION
**Pull request by vitorboschi**
This function is running git and also initializing a shell, which can be a relatively slow operation. By leveraging the jobs api, we run the command in background, reducing the time it takes for the buffer to be available to the user.

Fixes NvChad/ui#215

### DIFF
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -96,7 +96,7 @@ local default_plugins = {
       vim.api.nvim_create_autocmd({ "BufRead" }, {
         group = vim.api.nvim_create_augroup("GitSignsLazyLoad", { clear = true }),
         callback = function()
-          vim.fn.jobstart({"git", "-C", vim.fn.expand "%:p:h", "rev-parse"},
+          vim.fn.jobstart({"git", "-C", vim.loop.cwd(), "rev-parse"},
             {
               on_exit = function(_, return_code)
                 if return_code == 0 then
--- a/lua/plugins/init.lua
+++ b/lua/plugins/init.lua
@@ -96,13 +96,18 @@ local default_plugins = {
       vim.api.nvim_create_autocmd({ "BufRead" }, {
         group = vim.api.nvim_create_augroup("GitSignsLazyLoad", { clear = true }),
         callback = function()
-          vim.fn.system("git -C " .. '"' .. vim.fn.expand "%:p:h" .. '"' .. " rev-parse")
-          if vim.v.shell_error == 0 then
-            vim.api.nvim_del_augroup_by_name "GitSignsLazyLoad"
-            vim.schedule(function()
-              require("lazy").load { plugins = { "gitsigns.nvim" } }
-            end)
-          end
+          vim.fn.jobstart({"git", "-C", vim.fn.expand "%:p:h", "rev-parse"},
+            {
+              on_exit = function(_, return_code)
+                if return_code == 0 then
+                  vim.api.nvim_del_augroup_by_name "GitSignsLazyLoad"
+                  vim.schedule(function()
+                    require("lazy").load { plugins = { "gitsigns.nvim" } }
+                  end)
+                end
+              end
+            }
+          )
         end,
       })
     end,

### CODE REVIEW

**Comment by lucario387** [comment]
Don't put commit message with issue numbers or pr numbers, it will quickly lead to bunch of noises

**Comment by vitorboschi** [comment]
I have removed the reference from the commit msg. Should I remove it from the PR's description as well?
Also, could you elaborate a bit more on the noises problem? Linking commits to the issues that originated them are usually helpful when you need to dive into the context of a change.

**Comment by lucario387** [comment]
Pr descriptions are fine
For commits, no. Every time you forced push that commit to anywhere, be it to update something, it's merged to nvchad, or someone else forks nvchad, that commit link/pr link/person being tagged will add another "xx references this issue in their commit/pr", or that tagged person is pinged. This will be a lot of noises to those sometimes looking up on info, just to have to scroll through myriads of "references commits"

**Comment by vitorboschi** [comment]
That makes a lot of sense. Thanks for the explanation!

File: lua/plugins/init.lua [hunk 1]
```diff
@@ -96,13 +96,18 @@ local default_plugins = {
       vim.api.nvim_create_autocmd({ "BufRead" }, {
         group = vim.api.nvim_create_augroup("GitSignsLazyLoad", { clear = true }),
         callback = function()
-          vim.fn.system("git -C " .. '"' .. vim.fn.expand "%:p:h" .. '"' .. " rev-parse")
```
lucario387, line 99: Actually, why not just `vim.fn.system({"git", "-C", vim.fn.expand "%:p:h", "rev-parse"})`?

lua/plugins/init.lua
```diff
@@ -96,13 +96,18 @@ local default_plugins = {
...
-          if vim.v.shell_error == 0 then
-            vim.api.nvim_del_augroup_by_name "GitSignsLazyLoad"
-            vim.schedule(function()
-              require("lazy").load { plugins = { "gitsigns.nvim" } }
-            end)
-          end
+          vim.fn.jobstart("git -C " .. '"' .. vim.fn.expand "%:p:h" .. '"' .. " rev-parse",
```
lucario387: This still requires a shell, move the command to a list-like instead of a string

lucario387: See `:h jobstart()` and `:h system()`

lua/plugins/init.lua
```diff
@@ -96,13 +96,18 @@ local default_plugins = {
...
+            {
+              on_exit = function(_, return_code)
+                if return_code == 0 then
+                  vim.api.nvim_del_augroup_by_name "GitSignsLazyLoad"
+                  vim.schedule(function()
+                    require("lazy").load { plugins = { "gitsigns.nvim" } }
+                  end)
```
lucario387, line 106: Why schedule here again?

lua/plugins/init.lua
vitorboschi: I tried that way first, but was unable to retrieve the return code.

lua/plugins/init.lua
vitorboschi: It does, but as it's now async, it does not block the buffer anymore.

vitorboschi: But I'll try using a list instead. Maybe it is able to retrieve the return code

lua/plugins/init.lua
vitorboschi: Schedule runs the code on the main thread, while the job runs on an unspecified one, so to be safe, I kept the schedule call. I'm not sure it's actually necessary.

siduck: > Why schedule here again?

it was there before too, loading gitsigns after UI would be better

lua/plugins/init.lua
vitorboschi: So the list works fine as well on `jobstart`, so I'll use it, as it means we can skip the shell entirely.

siduck (approved):

PR #2587, opened 2024-01-02
Treesitter auto import

elferherrera:
Adding auto import to treesitter allows to have all required languages and not only lua. I was having issues reading help files because treesitter didn't have all the required tokens

### DIFF
--- a/lua/plugins/configs/treesitter.lua
+++ b/lua/plugins/configs/treesitter.lua
@@ -1,5 +1,5 @@
 local options = {
-  ensure_installed = { "lua" },
+  auto_install = true,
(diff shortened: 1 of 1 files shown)


siduck:
this you can add in your custom plugins  overrides

elferherrera:
How do you add overrides?

siduck:
@elferherrera https://nvchad.com/docs/config/plugins

nvchad uses lazy.nvim so basically user custom plugin tables will override default ones!

elferherrera:
Thanks. I found what I was looking for in the documentation

PR #2598, opened 2024-01-08
Update NvChad on <leader>cu

le4ker:
A mapping that allows to easily trigger the `NvChadUpdate` command.


siduck:
@le4ker this you can add in your custom mappings!

le4ker:
@siduck yeah [I've added it](https://github.com/le4ker/nvim-config/blob/v2.0/lua/custom/mappings.lua#L57), I just thought it might deserve an upstream commit as well so others can have it as well 😉

PR #2601, opened 2024-01-10
fix: ensure install all neovim bundled parsers

lucario387:
There can be more cases in the future where the bundled parser is not compatible with nvim-treesitter's queries, let's just ensure install all the bundled parsers to avoid the issue.

### DIFF
--- a/lua/plugins/configs/treesitter.lua
+++ b/lua/plugins/configs/treesitter.lua
@@ -1,5 +1,5 @@
 local options = {
-  ensure_installed = { "lua" },
+  ensure_installed = { "lua", "c", "vimdoc", "bash", "python", "markdown", "markdown_inline" },
 
   highlight = {
     enable = true,


siduck:
i can only include lua, vim, vimdoc by default. not others!

lucario387:
> i can only include lua, vim, vimdoc by default. not others!

Why then? All of the parsers listed above are bundled with neovim. I did forget that vim is also bundled

siduck:
not everyone uses bash python markdown, i dont want to favor one lang over other, only lua &vim&vimdoc must be added by default imo

lucario387:
This is not about "some will not use abc", this is about "ensuring if they use abc, they won't get any error"

siduck:
why tho, cant they    install it later

lucario387:
Because when it's incompatible, like how people are asking why :h is broken, just install the compatible version beforehand to lessen the amount of answers needed

siduck:
> Because when it's incompatible, like how people are asking why :h is broken, just install the compatible version beforehand to lessen the amount of answers needed

ive seen this only with vimdoc so far tho

lucario387:
> ive seen this only with vimdoc so far tho

Just because it only happens with one bundled parser, doesn't mean it won't happen to any other bundled parser.
Play safe

lucario387:
Closing since nvim-treesitter plans to handle this

PR #2610, opened 2024-01-17
Feature to toggle background transparency

Samar-exe:


siduck:
not needed. users can add it for their custom mappings!

PR #2621, opened 2024-01-20
Added binding Ctrl+q to quit

WinFan3672:
Considering that NvChad already has Ctrl+s, shouldn't it also have Ctrl+q?

### DIFF
--- a/lua/core/mappings.lua
+++ b/lua/core/mappings.lua
@@ -25,6 +25,7 @@ M.general = {
 
     -- save
     ["<C-s>"] = { "<cmd> w <CR>", "Save file" },
+    ["<C-q>"] = { "<cmd> q <CR>", "Close buffer" },
 
     -- Copy all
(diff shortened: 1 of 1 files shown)


siduck:
thats not simple! as I maintain our own bufferline plugin, use function used for <leader> x   mapping

WinFan3672:
> thats not simple! as I maintain our own bufferline plugin, use function used for x mapping

I would actually disagree; it works perfectly (it closes whatever is currently selected).

siduck:
> > thats not simple! as I maintain our own bufferline plugin, use function used for x mapping
> 
> I would actually disagree; it works perfectly (it closes whatever is currently selected).

it doesnt handle terminal buffers well, nor maintains the layout too and there are more edge cases, check https://github.com/NvChad/ui/blob/v2.0/lua/nvchad/tabufline/init.lua#L53

PR #2638, opened 2024-01-28
fix: Deep copy of key mappings instead of in-place modification

domwst:
In order to change default key mappings to the ones that I'm used to, I've tried to add something like this in my `custom/chadrc.lua`:

```lua
local default_mappings = require "core.mappings"

---@type ChadrcConfig
local M = {}

M.mappings = {
    disabled = {
        n = {
            ["<leader>ca"] = {},
        },
    },
    lspconfig = {
        n = {
            ["<leader>vca"] = default_mappings.lspconfig.n["<leader>ca"],
        },
    }
}
```

But because original mappings are modified in-place and (I suppose) because lua doesn't reload the `core.mappings` package but uses it's in-memory representation, I can't use the old mappings. If there is a proper way to remap the default key mappings, I'd be happy to know it. Thank you!

### DIFF
--- a/lua/core/utils.lua
+++ b/lua/core/utils.lua
@@ -16,7 +16,24 @@ M.load_config = function()
   return config
 end
 
-M.remove_disabled_keys = function(chadrc_mappings, default_mappings)
+-- Copied from http://lua-users.org/wiki/CopyTable
+M.deepcopy = function(orig)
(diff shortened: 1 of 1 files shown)


siduck:
is just a matter of time when this isnt needed. https://github.com/NvChad/NvChad/issues/2688

PR #2690, opened 2024-02-25
fix: Better approach to check if the system is Windows before setting PATH

pablojrl123:
Now the check for Windows no longer requires luv's function `os_uname()`, which might not have a unique value when neovim is built using different toolchains (MSVC, MinGW, etc).

For example, the neovim version distributed in the `mingw-w64-ucrt-x86_64-neovim` package from MSYS2 reports itself as `MINGW32_NT-10.0` rather than `Windows_NT`, which causes PATH to not be set properly. This has been fixed.

### DIFF
--- a/lua/core/init.lua
+++ b/lua/core/init.lua
@@ -57,7 +57,7 @@ for _, provider in ipairs { "node", "perl", "python3", "ruby" } do
 end
 
 -- add binaries installed by mason.nvim to path
-local is_windows = (vim.fn.has("win32") ~= 0)
+local is_windows = vim.fn.has("win32") ~= 0
 vim.env.PATH = vim.fn.stdpath "data" .. "/mason/bin" .. (is_windows and ";" or ":") .. vim.env.PATH
 
 -------------------------------------- autocmds ------------------------------------------
--- a/lua/core/init.lua
+++ b/lua/core/init.lua
@@ -57,7 +57,7 @@ for _, provider in ipairs { "node", "perl", "python3", "ruby" } do
 end
 
 -- add binaries installed by mason.nvim to path
-local is_windows = vim.loop.os_uname().sysname == "Windows_NT"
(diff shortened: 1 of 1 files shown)


siduck:
image

i dont see how its better approach other than shorter syntax :thinking:

pablojrl123:
> i dont see how its better approach other than shorter syntax 🤔

Fair enough, though read what I said about the different possible values of `os_uname().sysname` in the PR. We'll need to check if it is `Windows_NT` or `MINGW32_NT-*`. Then again, I don't know if NvChad wishes to support MSYS2, feel free to disregard if that's not the case.

siduck:
> MSYS2

i honestly dont have idea about windows, all systems in my home use linux :grin:  so i wouldnt even know its basics too!

siduck:
@lucario387 could help

lucario387:
https://github.com/neovim/neovim/blob/268066e01400f55b0c38716f0d6ee3dece10c43e/test/helpers.lua#L375

Yes we should. I'd prefer a `sysname:lower():match("mingw") or sysname:lower():match("windows")` like neovim. So we can blame neovim if things go south later :)

ppinkhasov (approved):

siduck:
whats wrong with current one tho? it works :thinking:

lua/core/init.lua
```diff
@@ -57,7 +57,7 @@ for _, provider in ipairs { "node", "perl", "python3", "ruby" } do
 end
 
 -- add binaries installed by mason.nvim to path
-local is_windows = vim.loop.os_uname().sysname == "Windows_NT"
+local is_windows = (vim.fn.has("win32") ~= 0)
```
siduck: i dont think (  ) is needed, format that file and see

siduck (approved):

PR #2715, opened 2024-03-10
add vim.g.vscode_snippets_exclude option

wSedlacek:
The `exclude` option for luasnip is useful for disabling snippets which a user may not want on by default.

For example the `all` option includes a snippet for `timeHMS` which always gets caught by `cmp` when typing `ts` for me which means all my typescript files end up as `file.16:07:21`

### DIFF
--- a/lua/nvchad/configs/luasnip.lua
+++ b/lua/nvchad/configs/luasnip.lua
@@ -1,5 +1,5 @@
 -- vscode format
-require("luasnip.loaders.from_vscode").lazy_load()
+require("luasnip.loaders.from_vscode").lazy_load { exclude = vim.g.vscode_snippets_exclude or {} }
 require("luasnip.loaders.from_vscode").lazy_load { paths = "your path!" }
 require("luasnip.loaders.from_vscode").lazy_load { paths = vim.g.vscode_snippets_path or "" }
 


siduck:
i think you could add that in your luasnip config as well

wSedlacek:
I have attempted adding `require("luasnip.loaders.from_vscode").lazy_load { exclude = {"all"} }` but sadly it doesn't overwrite the existing one that has been loaded.

Sadly this isn't an option on the plugin's setup (otherwise I could just add it to the lazy config and have it merged in)

siduck:
did you add it in your config function? show me what u did!

wSedlacek:
Here is an attempt in the `config` for `luasnip`.
https://github.com/wSedlacek/nvim/blob/05b75ae1e2ef9c44bf31ca34ca6e7ab1f3e28388/lua/plugins/luasnip.lua#L7

I have tried it both before and after the loading of the nvchad config.
Of course I could disable the nvchad config for luasnip, but I would prefer not to maintain the autocmd and the rest of the configuration if possible.

siduck:
weird, ig it didnt work cuz nvim caches lua modules. Wont suggest to use plenary to reload that module lol

siduck:
Also i saw your nvchad config.

image


these must be loaded in your main init.lua before it was in custom/init.lua. You could load them in a file and then load that file in main init.lua to keep it tidy.

Also extended_integrations is removed, check https://nvchad.com/docs/config/theming#integrations

wSedlacek:
Thank you very much for your feedback! I have updated my config.
I noticed that the types aren't updated yet, will those be coming soon?

https://github.com/NvChad/ui/blob/83f06c444020f619a91a20d8cb350b59eb1ca1cd/nvchad_types/chadrc.lua#L60

siduck:
@wSedlacek idk how lua types work, some contributor maintained it. Probably after april 11 i'll learn how types work and add it

siduck:
if you want to contribute with types then you can do it :grimacing:

wSedlacek:
I was thinking I might have a look at it but wasn’t gonna commit to anything.
You’ll see a PR if I get something working! Thank you again for all your hard work!

PR #2729, opened 2024-03-13
add more gitsigns mappings

KorigamiK:
this adds `[c` ,`]c`, `<leader>td` mappings like in v2.0


siduck:
I want to have only very few gitsigns mappings by default  as they're a lot , you're free to add in your config

KorigamiK:
> I want to have only very few gitsigns mappings by default  as they're a lot , you're free to add in your config

This should be a breaking change for those updating from v2

PR #2743, opened 2024-03-18
feat: restore cursor position from last time

Groveer:


siduck:
this feature many might not like, ig you can keep it in your config. I'll add it into the recipes section of docs

https://nvchad.com/docs/recipes

gatspy:
> this feature many might not like, ig you can keep it in your config. I'll add it into the recipes section of docs
> 
> https://nvchad.com/docs/recipes

I added it in the my config, but it didn't work. why?

siduck:
@gatspy works here, where did you add it?

PR #2771, opened 2024-04-04
New option "ui.tabufline.show_tabs_always"

thenullvek:
Draw the tabs even if there's only one buffer opened.


siduck:
by default when there's just 1 buffer opened, tabufline will not     load, once u open one more buffer, it'll load.     Now if u close all buffers and open just 1 buffer then tabufline will still say, so basically tabufline only once will lazyload like that

in your chadrc `lazyload=false` 

And i dont think its beneficial to show the "tabs" switcher if there's just 1 tab as we already know :))

PR #2830, opened 2024-05-09
Added binds for moving lines of code

kuprTheMan:
I really like the feature from Vs Code that allows you to use the keys on your keyboard to move pieces of code relative to the rest of the code around, I think it's a handy feature that will help well when editing code

### DIFF
--- a/lua/nvchad/mappings.lua
+++ b/lua/nvchad/mappings.lua
@@ -51,6 +51,38 @@ map("n", "<leader>/", function()
   require("Comment.api").toggle.linewise.current()
 end, { desc = "comment toggle" })
 
+--Move line
+local function move_line(op)
+	return function()
+		local start = op == "+" and 1 or 2
+		local count = vim.v.count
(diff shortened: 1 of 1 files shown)


YagaYasha:
Always liked that feature. Sounds cool

siduck:
it could go in   your mappings file!! honestly i would like this feature but the amount of lua function needed for this makes me sad!

Alexis12119:
Something Like this?
```lua
vim.keymap.set("n", "<A-j>", "<cmd>m .+1<CR>==", { desc = "move the line up", silent = true })

vim.keymap.set("n", "<A-k>", "<cmd>m .-2<CR>==", { desc = "move the line down", silent = true })

vim.keymap.set("v", "<A-j>", "<cmd>m '>+1<CR>gv=gv", { desc = "move the selected text up", silent = true })

vim.keymap.set("v", "<A-k>", "<cmd>m '<-2<CR>gv=gv", { desc = "move the selected text down", silent = true })
```

PR #2833, opened 2024-05-11
Fix a conflict keymapping

wangrecv:
The gi keymapping which triggers `vim.lsp.buf.implementation` is conflicted with an existed keymapping which descripted by which-key as "Move to the last insertion and INSERT".

I suggest replace the keymapping to `<leader>i`.


siduck:
it is the default config suggested by     lspconfig and removing it will be a breaking change. So i suggest you to disable for yourself

Issue #2796, opened 2024-04-19
I can't assign the toggle comment function to the gc command.


kamijodev:
**Describe the bug**

Hello! I recently started using nvchad as a beginner.
I want to execute comment toggling with gc, but it behaves strangely. Sometimes which-key.nvim appears.
Is this a bug, or is there something wrong with my settings?

It might be similar to #1972 

**To Reproduce**

```lua
nomap("n", "gc")
nomap("n", "gcc")

vim.keymap.set('n', 'gc',
  function()
    require("Comment.api").toggle.linewise.current()
  end, { desc = "Comment Toggle" }
)
```

**Desktop (please complete the following information):**
 - Operating System: Endeavour OS
 - Terminal: Alacritty
 - Version of Neovim: v0.10.0

siduck:
![image](https://github.com/NvChad/NvChad/assets/59060246/a6af181f-c41b-46ef-844d-8e8c3d922874)

i guess you have to disable here

siduck:
so in your custom plugins add that comment table ( not all of its options ), only keys and the mappings you need

PR #2861, opened 2024-05-29
comments: use neovim native commenting

lianghx-319:
I think nvchad can use neovim native commenting to keep the core clear & tiny.

User can choose other commenting plugin like `numToStr/Comment.nvim` or `folke/ts-comments.nvim`

### DIFF
--- a/lua/nvchad/mappings.lua
+++ b/lua/nvchad/mappings.lua
@@ -44,15 +44,8 @@ map("n", "<leader>x", function()
 end, { desc = "buffer close" })
 
 -- Comment
-map("n", "<leader>/", function()
-  -- https://github.com/neovim/neovim/blob/efa45832ea02e777ce3f5556ef3cd959c164ec24/runtime/lua/vim/_defaults.lua#L142
-  return require("vim._comment").operator() .. "_"
-end, { desc = "comment toggle", expr = true })
-
-map("v", "<leader>/", function()
-  -- https://github.com/neovim/neovim/blob/efa45832ea02e777ce3f5556ef3cd959c164ec24/runtime/lua/vim/_defaults.lua#L137
-  return require("vim._comment").operator()
-end, { desc = "comment toggle", expr = true })
+map("n", "<leader>/", "gcc", { desc = "comment toggle", remap = true })
+map("v", "<leader>/", "gc", { desc = "comment toggle", remap = true })
 
 -- nvimtree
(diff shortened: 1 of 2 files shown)


siduck:
i was about to remove it in few days but thought  about adding ts-comments

siduck:
but then thought its only usecase could be for tsx/jsx :thinking:  https://github.com/folke/ts-comments.nvim/blob/main/lua/ts-comments/config.lua

And we dont cater any specific language settings in nvchad except for lua. 

But i still miss the leader / keybind! hopefully we can map it

lianghx-319:
@siduck I have binding `leader /` back, but using the internal api may has changed risk.

siduck:
i  think we can just map them to keys

```lua

map("n", "<leader>/", "gcc", { desc = "comment toggle", remap = true })
map("v", "<leader>/", "gc", { desc = "comment toggle", remap = true })
```

siduck (approved):

lianghx-319:
> i  think we can just map them to keys
> 
> 
> 
> ```lua
> 
> 
> 
> map("n", "<leader>/", "gcc", { desc = "comment toggle", remap = true })
> 
> map("v", "<leader>/", "gc", { desc = "comment toggle", remap = true })
> 
> ```

got new option,I just noob of neovim & lua

siduck:
> got new option,I just noob of neovim & lua

All good, I like such PR's!

image

MayDomine:
Hey guys，space+/ does not work anymore because this PR recently.
This  because my nvim version is outdated. 
Maybe I should disable nvchad update?
Can you guys tell me how to do it?
I will appreciate it.

siduck:
@MayDomine just update neovim, most distros or macos probably have it 0.10 i think, or else use bob neovim version manager which is the best cross platform tool for managing neovim versions :) 

or just use the binary from the neovim releases page

or lock nvchad plugin, which is in your main init.lua file

```lua
 { 
     "NvChad/NvChad",
     commit = "0496016e188e294ba09dde41914b83800a9fa9ca"
  }
```

KorigamiK:
So, block comments are gone now? What's the best way to bring them back? @siduck

KorigamiK:
I think this is good enough
https://github.com/NvChad/NvChad/issues/2796#issuecomment-2066519568

siduck:
> So, block comments are gone now? What's the best way to bring them back? @siduck

just use comment.nvim

PR #2867, opened 2024-06-01
fix(MasonInstallAll):  mason registry refresh callback

saifulapm:
Added refresh callback to make sure install package on first command. Right now it's not installing on first command.


siduck:
it installs on first command here

siduck:
https://github.com/NvChad/NvChad/assets/59060246/a230236b-c999-4b2f-a98e-82be67782180

saifulapm:
https://github.com/NvChad/NvChad/assets/3833316/b2c4b179-435d-4c37-a922-ba502da28923

If you delete ~/.local/share/nvim, I mean if you install NvChad as new user, you will see this issue.

PR #2886, opened 2024-06-10
feat(ui) Set default terminal signcolumn as active

haistss:
[feat(ui) Added option to the terminal signcolumn as padding #292](https://github.com/NvChad/ui/pull/292)


siduck:
not needed anymore! cuz one can just add those win options in winopts table now

https://github.com/NvChad/NvChad/commit/0bb40c98062f5dc788ebf6da1d1336b8161c5885

PR #2940, opened 2024-07-24
performance: defer clipboard because xsel and pbcopy can be slow

abeldekat:
See [this](https://github.com/LazyVim/LazyVim/discussions/4112) discussion in LazyVim

The approach discussed has been merged in [LazyVim](https://github.com/LazyVim/LazyVim/pull/4120) and [AstroNvim](https://github.com/AstroNvim/astrocore/pull/28)

TLDR: 
Startup time performance is affected quite significantly when the clipboard provider is `xsel`(linux) or `pbcopy`(macos). I expect an improvement in these cases, especially on older pc's.

This PR makes sure that `vim.opt.clipboard` is only applied after `UiEnter`

@siduck, in case you are interested, please let me know if the code needs adjustments.
I "decorated" the config function defined in `Nvchad/starter".

### DIFF
--- a/lua/nvchad/options.lua
+++ b/lua/nvchad/options.lua
@@ -1,3 +1,16 @@
+local group = vim.api.nvim_create_augroup("NvLazyClipboard", { clear = true })
+vim.api.nvim_create_autocmd("User", {
+  group = group,
+  pattern = "LazyDone",
+  callback = function()
+    local lazy_clipboard = vim.opt.clipboard -- could be changed by the user
+    vim.opt.clipboard = ""
+    vim.schedule(function() -- defer setting clipboard
+      vim.opt.clipboard = lazy_clipboard
+    end)
+  end,
+})
+
 local opt = vim.opt
 local o = vim.o
 local g = vim.g
--- a/lua/nvchad/plugins/init.lua
+++ b/lua/nvchad/plugins/init.lua
@@ -1,6 +1,26 @@
 return {
 
   {
+    "NvChad/NvChad",
+    init = function()
+      local spec = require("lazy.core.config").spec.plugins["NvChad"]
+      if not spec.config then
+        return
+      end
+
+      local config_function = spec.config -- NvChad/starter: requires options.lua
+      spec.config = function(...)
+        config_function(...)
+        local lazy_clipboard = vim.opt.clipboard -- defer setting clipboard
+        vim.opt.clipboard = ""
+        vim.schedule(function()
+          vim.opt.clipboard = lazy_clipboard
+        end)
+      end
+    end,
+  },
+
+  {
     "NvChad/base46",
     branch = "v2.5",
     build = function()
--- a/lua/nvchad/plugins/init.lua
+++ b/lua/nvchad/plugins/init.lua
@@ -1,26 +1,6 @@
 return {
 
   {
-    "NvChad/NvChad",
-    init = function()
-      local spec = require("lazy.core.config").spec.plugins["NvChad"]
-      if not spec.config then
-        return
-      end
-
-      local config_function = spec.config -- NvChad/starter: requires options.lua
-      spec.config = function(...)
-        config_function(...)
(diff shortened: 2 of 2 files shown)


siduck:
the code doesnt look satisfying to me. it must be added in the starter :/

siduck:
https://github.com/NvChad/NvChad/blob/v2.5/lua/nvchad/options.lua#L12

you should add it here

abeldekat:
> https://github.com/NvChad/NvChad/blob/v2.5/lua/nvchad/options.lua#L12
> 
> you should add it here

Unfortunately, if the user decides to change the clipboard option in his own lua/options.lua, there is no way to reset and defer the option.

-- User options.lua in "NvChad/starter":

```lua
require "nvchad.options"
-- add yours here!

local o = vim.o
-- o.cursorlineopt ='both' -- to enable cursorline!

-- Not deferred!
o.clipboard = "unnamed" -- or a custom clipboard tool config, see ":h g:clipboard"
```

-- Nvchad options.lua in "NvChad/NvChad":

```lua
vim.schedule(function()
  o.clipboard = "unnamedplus" -- overrides user's changes...
end)
```

I also tried the approach in an autocommand on [`LazyLoad`](https://lazy.folke.io/usage#-user-events), added as the first statement to lua/nvchad/plugins/init.lua. That did not improve the performance, probably kicked in too late.

EDIT: An example of "doing something" with spec.config supplied by the user can be found [here](https://github.com/LazyVim/LazyVim/blob/71b26905822e81ec4a3842947d32a1b46c185948/lua/lazyvim/plugins/formatting.lua#L58)

siduck:
ig its better to remove the cliboard function from main repo itself and let the user add it in his starter..

I dont get why such simple option should be heavy, i hope its not cuz of windows :fearful:  :rage:

abeldekat:
> ig its better to remove the cliboard function from main repo itself and let the user add it in his starter..

Yes, that would be the best option. I think that existing users would not profit though.

> I dont get why such simple option should be heavy, i hope its not cuz of windows 😨 😡

That's discussed [here](https://github.com/LazyVim/LazyVim/discussions/4112). It can happen on linux("xsel"), macos("pbcopy") and windows-wsl2, as far as I can see.

Would you like me to propose a PR on the starter?

siduck:
what if the main nvchad's repo's option file had just this?

```lua
vim.schedule(function()
  o.clipboard = "unnamedplus" -- overrides user's changes...
end)
```

siduck:
what if the main nvchad's repo's option file had just this?

```lua
vim.schedule(function()
  o.clipboard = "unnamedplus" -- overrides user's changes...
end)
```

abeldekat:
That would work except when the user adds his own clipboard setting to `lua/options.lua`. The user could decide to opt-out (""), or change ("unnamed" for example) or provide a completely custom config according to `:h g:clipboard`. As the default `vim.o.clipboard` option also has the empty string as value, it is not trivial to distinguish between user-changes and that default.

The problem, I think, is that in the starter, in NvChad's config function, `lua/options.lua` is required. The effect is that the loading of the options is "in control" by the user, not by NvChad.

That's why I ended up with `decorating` that config function, taking back control. I think that works in 99.99% of the cases. However, because the user is free to put any code he desires inside the config function, the "timing" could still break. Highly unlikely in my opinion.

abeldekat:
@siduck,

I found a better approach and pushed a new commit. I used the local `lua/options.lua` to test:

1. No override
2. `o.clipboard = ""`
3. `o.clipboard = "unnamed"
4. `vim.opt.clipboard = ""`

The code now uses an `autocommand` on `LazyDone`: "when lazy has finished starting up and loaded your config"

siduck:
@abeldekat the user can add his own vim.schedule... in his options file

abeldekat:
Yes, the user can do:

```lua
o.clipboard = "" -- reset
vim.schedule(function()
  o.clipboard = "unnamedplus"
end)
```

PR #3036, opened 2024-10-10
whichkey: set event to VeryLazy

Pairadux:
Fixes issue where the first instance of <leader> press is ignored.

### DIFF
--- a/lua/nvchad/plugins/init.lua
+++ b/lua/nvchad/plugins/init.lua
@@ -57,6 +57,7 @@ return {
 
   {
     "folke/which-key.nvim",
+    event = "VeryLazy",
     keys = { "<leader>", "<c-r>", "<c-w>", '"', "'", "`", "c", "v", "g" },
     cmd = "WhichKey",
     opts = function()


siduck:
it breaks the point of lazyloading whichkey. iirc it used to work on first press of keys before, until the new version of whichkey was released

Pairadux:
I will look into whichkey and see if I can figure out a way that allows lazy loading while fixing this issue, will update my PR accordingly.

Pairadux:
The official documentation for Which-Key says to use event="VeryLazy", idk if this is just now a thing or not with V3.0, but it seems to be how it should be implemented.

siduck:
@folke before v3.0 of whichkey, it was possible to lazyload whichkey on keys, like in the version v2.30

when the user pressed `<leader>`  whichkey would get lazy-loaded

but now when the user presses it, whichkey doesnt popup unless he presses for the second time. Is this a bug or normal

siduck:
any idea? @folke :fearful:

PR #3039, opened 2024-10-12
Add resume previous search with telescope

Sumanthbhumi:
Continue the previous search with telescope from where we left it.

I know users can add it themselves, but this is a great feature that should come with NvChad and doesn't conflict with other keybinds.


siduck:
<leader>f sounds bad, i'd try leader r

Sumanthbhumi:
added  with `f` because most search with telscope started with `f` so i thought it would be more convenient

siduck:
thats subjective or you mean most telescope mappings of ours start with f?

Sumanthbhumi:
i meant nvchad mappings

PR #3054, opened 2024-10-18
Add missing menu mappings

PedroClerici:


siduck:
I want to wait for 1-2 months for menu plugin to become stable and let the users test it, then I will add it.

PR #3057, opened 2024-10-20
fix(telescope): lazy load extensions in defined list

dangdd2003:
Should we really need to remove auto load extensions function ? I'm using `file_browser` extensions, and I saw it failed to load my config after newest commit.

Up to now, we can remove `terms` and `themes`because they work independently.

### DIFF
--- a/lua/nvchad/configs/telescope.lua
+++ b/lua/nvchad/configs/telescope.lua
@@ -19,6 +19,6 @@ return {
     },
   },
 
-  extensions_list = { "themes", "terms" },
+  extensions_list = {},
   extensions = {},
 }
--- a/lua/nvchad/plugins/init.lua
+++ b/lua/nvchad/plugins/init.lua
@@ -152,6 +152,14 @@ return {
     opts = function()
       return require "nvchad.configs.telescope"
     end,
+    config = function(_, opts)
+      local telescope = require "telescope"
+      telescope.setup(opts)
+      -- load extensions
+      for _, ext in ipairs(opts.extensions_list) do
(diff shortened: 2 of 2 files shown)


siduck:
i dont think this is  needed. `Telescope terms or themes` command just works.

siduck:
thats why i removed it before, and users would just map it to something or use the command

dangdd2003:
So we should remove the list. It is really confusing when we defined an extension list without calling ‘require’.

siduck:
image


remove the loading of the extensions too

siduck:
idk why it shows as add diff here even tho we already have that code :thinking:

dangdd2003:
I'm still very confused here, why do we need that `extension_list` even though when i delete it, it still works fine with two command `Telescope terms and themes` ? You have already registered them [here](https://github.com/NvChad/ui/tree/v3.0/lua/telescope/_extensions).

I saw you removed that [part](https://github.com/NvChad/NvChad/commit/7f055761ba0d94a562ead5ff85352c4c67b51f67#diff-56da2ecc4b8e5c3ad4fa01140276f112365f869793e18756ac8dc45dee28953aL166-L175) of code, so NVChad cannot auto load my config for `file_browser` extension.
image

siduck:
we needed them so it'll show in Telescope <> command autocompletion in cmdline and it will show in :Telescope window itself

however i dont want that anymore

siduck:
Ok so in your PR, just remove these

image

dangdd2003:
Thanks! For the clean commits, i will close this and open new pull rq.

PR #3095, opened 2024-11-11
Add theme toggle mapping

voidanix:
Currently, it seems cycling between the themes is only possible when the tabs line is visible. Add a mapping to quickly switch between the themes specified in chadrc.


siduck:
you can add it in your own mapping config.

PR #3119, opened 2024-11-29
Add ufo plugin. Used for folding

Kennedy242:
See it on GitHub: https://github.com/kevinhwang91/nvim-ufo

The config for the demo is used

I think this addition would be useful to lots of users of NvChad. Especially if they are coming from VSCode where code folding was built in. 

> I know that in general adding more plugins is frowned upon...

I will update the NvChad docs if this is seen as a positive addition


siduck:
users can add it in their configs

PR #3178, opened 2025-01-24
feat(mappings): add Ctrl + Backspace support to delete a word in insert mode

bvsvntv:
I've added a keymap to make Vim more accessible to newcomers by introducing support for deleting a word using `Ctrl + Backspace` in insert mode. This behavior is common in other editors and enhances the onboarding experience for users transitioning to Vim.

The keymap maps `<C-BS>` to Vim's built-in `<C-w>` command, which is already familiar to experienced Vim/Neovim users for deleting the previous word. 

Even if this functionality already exists, I believe someone will find it helpful ;)

Both `<C-BS>` and `<C-H>` are included to ensure compatibility, as key interpretation can vary across terminals, but either keymap can be removed accordingly based on the terminal's behavior.
```lua
map("i", "<C-BS>", "<C-w>", { desc = "general delete the previous word in insert mode" })
map("i", "<C-H>", "<C-w>", { desc = "general delete the previous word in insert mode" })
```

### More on key interpretation

We can check our terminal's interpretation by running `cat -v` in our terminal and pressing `Ctrl + Backspace`. 
> For example, If we see `^H` (`^` represents `Ctrl`), our terminal supports `<C-H>` version. In this case, we can remove the `<C-BS>` keymap.

### DIFF
--- a/lua/nvchad/mappings.lua
+++ b/lua/nvchad/mappings.lua
@@ -17,6 +17,9 @@ map("n", "<Esc>", "<cmd>noh<CR>", { desc = "general clear highlights" })
 map("n", "<C-s>", "<cmd>w<CR>", { desc = "general save file" })
 map("n", "<C-c>", "<cmd>%y+<CR>", { desc = "general copy whole file" })
 
+map("i", "<C-BS>", "<C-w>", { desc = "general delete the previous word in insert mode" })
+map("i", "<C-H>", "<C-w>", { desc = "general delete the previous word in insert mode" })
+
 map("n", "<leader>n", "<cmd>set nu!<CR>", { desc = "toggle line number" })
 map("n", "<leader>rn", "<cmd>set rnu!<CR>", { desc = "toggle relative number" })
 map("n", "<leader>ch", "<cmd>NvCheatsheet<CR>", { desc = "toggle nvcheatsheet" })


siduck:
why just not backspace? The plugin will just use backspace. its the standard everywhere

bvsvntv:
> why just not backspace? The plugin will just use backspace. its the standard everywhere

I understand your point about using Backspace, as it is the standard key for deleting characters. However, my intent with this mapping was to replicate the behavior of `Ctrl + Backspace`, which is specifically used in many editors (e.g., VS Code, Sublime, IntelliJ) to delete entire words.

siduck:
> > why just not backspace? The plugin will just use backspace. its the standard everywhere
> 
> I understand your point about using Backspace, as it is the standard key for deleting characters. However, my intent with this mapping was to replicate the behavior of `Ctrl + Backspace`, which is specifically used in many editors (e.g., VS Code, Sublime, IntelliJ) to delete entire words.

I'll expose an option where user can add mappings to the typr window buffer!

siduck:
oh my bad, i misread. i was thinking this pr was in my typr plugin. Anways you can add it in your config's mapping file. I will not add more mappings to nvchad main repo now unless its highly important

bvsvntv:
Oh! OK.
I've been using this mapping in my own Neovim config for quite a while, and recently I wanted to try out some derivatives of Neovim. I’ve heard a lot of great things about NvChad, especially about its UI, so I decided to give it a shot, and I really like the UI. Since I was already familiar with my mappings, I tried using `Ctrl + Backspace` to delete an entire word, but it didn't work. So, I skimmed through the code base and added these mappings. I thought it might be helpful for newcomers, so I made the PR.

I think **not adding too many mappings to the main repo unless it's really necessary** is a great decision. Things can get pretty messy when everyone submits PRs with their own mapping preferences, like I did. HEHE ;)

PS: By the way, you have a really nice and cool project! Keep up the great work.

siduck:
@iambasantarai  put that mapping in your mappings file, if it doesn't work there then I'll debug

bvsvntv:
It works.

PR #3180, opened 2025-01-31
Add auto_install to Treesitter to automatically install missing parsers

dodofarm:
Just thought it might be a great idea to add this functionality, I can't see any downsides and think many might benefit.

Correct me if I'm wrong though.

https://github.com/nvim-treesitter/nvim-treesitter/blob/7e3942ceca9e0c28760f77ac33bc16399146d879/doc/nvim-treesitter.txt#L55

### DIFF
--- a/lua/nvchad/configs/treesitter.lua
+++ b/lua/nvchad/configs/treesitter.lua
@@ -6,6 +6,8 @@ end)
 return {
   ensure_installed = { "lua", "luadoc", "printf", "vim", "vimdoc" },
 
+  auto_install = true,
+  
   highlight = {
     enable = true,
     use_languagetree = true,


siduck:
i hesitated about this before as i'm not sure if treesitter plugin makes the check on startup , if it does then it might cost minuet startuptime

dodofarm:
It is indeed called on setup:
https://github.com/nvim-treesitter/nvim-treesitter/blob/48f7ce34ef274af10a3b2aa24932a310b1bd055d/lua/nvim-treesitter/configs.lua#L418

Right before the "ensure_installed" check.

Here's the function that's called when the setting is enabled:
https://github.com/nvim-treesitter/nvim-treesitter/blob/48f7ce34ef274af10a3b2aa24932a310b1bd055d/lua/nvim-treesitter/install.lua#L557

As far as I can see it mainly does:
 - Getting buffer language
 - Checking if parser exists (hashmap lookup)
 - Checking if already installed (simple file check)

Doubt it should impact performance enough to outweigh the convenience of not having to maintain a custom "ensure_installed". Other distros (e.g. LazyVim) seem to individually trigger treesitter with "ensure_installed" for each specific filetype. e.g. [this](https://github.com/LazyVim/LazyVim/blob/29829f7eb6e937f9fe43e9455e55704b3fac8794/lua/lazyvim/plugins/extras/lang/ruby.lua#L23)

But I can see your point, feel free to not merge this if you think the performance impact is considerable enough and the users can maintain it themselves.

From my personal checking I didn't notice higher startup times in the Lazy profiler.

siduck:
> From my personal checking I didn't notice higher startup times in the Lazy profiler.

i will test this on my old pentium machine

sayantachaudhary:
> i hesitated about this before as i'm not sure if treesitter plugin makes the check on startup , if it does then it might cost minuet startuptime

        auto_install = vim.fn.executable("git") == 1 and vim.fn.executable("tree-sitter") == 1,

siduck:
> > i hesitated about this before as i'm not sure if treesitter plugin makes the check on startup , if it does then it might cost minuet startuptime
> 
> ```
>     auto_install = vim.fn.executable("git") == 1 and vim.fn.executable("tree-sitter") == 1,
> ```

this is another headache, it'll run that function! the user can add it himself

topphh (approved):

PR #3265, opened 2025-04-24
fix(lspconfig): use more stable 0.11 compatibility check

anthraxx:
Previously only the minor version has been checked for equality, which means it is brittle whenever we reach 0.12. Independently of when this is expected to be removed, it's simply much safer to check if the expected functionality is available by testing `vim.lsp.enable` for nil.

Fixes 84f8170fdab2a9bb82f36fe76c68bca2241c50d8

### DIFF
--- a/lua/nvchad/configs/lspconfig.lua
+++ b/lua/nvchad/configs/lspconfig.lua
@@ -72,7 +72,7 @@ M.defaults = function()
 
   -- Support 0.10 temporarily
 
-  if vim.version().minor == 11 then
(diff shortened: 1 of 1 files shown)


siduck:
just check for vim.lsp.config then, why .enable?

anthraxx:
Guess i just wanted to be extra cautious, i'll remove .enable

anthraxx:
adjusted, only checking for `vim.lsp.config` now

PR #3401, opened 2026-04-18
Replace command based keymaps with lua api calls

kzzzld:
I replaced command based keymaps (like :Telescope find_files) with lua apis (like require"telescope.builtin").

### DIFF
--- a/lua/nvchad/mappings.lua
+++ b/lua/nvchad/mappings.lua
@@ -54,27 +54,55 @@ map("n", "<C-n>", "<cmd>NvimTreeToggle<CR>", { desc = "nvimtree toggle window" }
 map("n", "<leader>e", "<cmd>NvimTreeFocus<CR>", { desc = "nvimtree focus window" })
 
 -- telescope
-map("n", "<leader>fw", "<cmd>Telescope live_grep<CR>", { desc = "telescope live grep" })
-map("n", "<leader>fb", "<cmd>Telescope buffers<CR>", { desc = "telescope find buffers" })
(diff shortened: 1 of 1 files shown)


siduck:
why?

UmutKzl:
@siduck First of all, thanks for answering.

I think it's more future-proof because Neovim development is centered in lua. Most plugins and distros prefer using lua because of that.

siduck:
> @siduck First of all, thanks for answering.
> 
> I think it's more future-proof because Neovim development is centered in lua. Most plugins and distros prefer using lua because of that.

yes but i think they will not remove those commands

UmutKzl:
> > @siduck First of all, thanks for answering.
> > I think it's more future-proof because Neovim development is centered in lua. Most plugins and distros prefer using lua because of that.
> 
> yes but i think they will not remove those commands

Yeah, that makes sense. My point wasn’t about commands being removed, but more about reducing abstraction layers and sticking closer to plugin APIs directly.
It might not be a functional improvement, more of a consistency/style preference with Lua-first Neovim development.

siduck:
you're somewhat right but its increasing the config loc for no better reason! and people remember the commands. their commands will never change ive been using them sicne ages. 

lua api makes sense if you want to configure more than what the default commands offer

UmutKzl:
> you're somewhat right but its increasing the config loc for no better reason! and people remember the commands. their commands will never change ive been using them sicne ages.
> 
> lua api makes sense if you want to configure more than what the default commands offer

Fair point but users will set their keymaps from mappings.lua and they can use commands when they want. Developers will know how to use Lua apis, so I think there won't be a problem. If you think there would be problems, I completely respect it.

PR #3409, opened 2026-06-15
fix(mappings): tab expanding snippets

Dragit:
when entering insert mode on top of a snippet name, pressing <Tab> would cause the snippet to expand, without any visual queues about this behaviour. changed to reflect recommended config from nvim-cmp: https://github.com/hrsh7th/nvim-cmp/wiki/Example-mappings#luasnip

### Example

When insert mode is entered on something which is identical to a LuaSnippet trigger, like `--`, for the snippet `region~` in Lua, pressing `<Tab>` expands the snippet.

In the example below insert mode was entered with `a` on a lua file containing `--`, which results in the cursor at the position of the `+`

```lua
--+
```

As no cmp menu is shown when entering insert mode, the result of pressing `<Tab>` is:

```lua
--#region
```

However the expected result would be:

```lua
--<Tab>
```

With `<Tab>` as the chars configured to be placed when pressing the tab key.
This is precisely what my changes accomplish.

### DIFF
--- a/lua/nvchad/configs/cmp.lua
+++ b/lua/nvchad/configs/cmp.lua
@@ -27,8 +27,8 @@ local options = {
     ["<Tab>"] = cmp.mapping(function(fallback)
       if cmp.visible() then
         cmp.select_next_item()
-      elseif require("luasnip").expand_or_jumpable() then
-        require("luasnip").expand_or_jump()
+      elseif require("luasnip").locally_jumpable(1) then
+        require("luasnip").jump(1)
       else
         fallback()
       end


siduck:
hmm show me video with/without. We have that function since ages!

Dragit:
Ok sorry tested it on a clean setup and it was fixed. Something in my config messed things up....
Very sorry for wasting your time.

siduck:
check the plugins you have

Dragit:
Nvm, I managed to reproduce it with a clean install!
here is the video proof:

https://github.com/user-attachments/assets/e527df05-6aca-4bbb-b181-99810fd20b35

Dragit:
Should I add a Video with my change implemented as well?
But it is pretty simple:
Instead of the snippet expanding there is just insertion of tab...
I thought this is the expected behavior if there is no completion menu visible.

siduck:
i'll check this on sunday! sorry for the late review :((

